PR #18, opened 2020-05-05
Alteration of currently_active_shift_for_venue

### LINKED ISSUE #7: Shift admin must be available after shift
The home page must display a button to go to the shift admin of a just ended shift. Now it displays "Start new shift" when the shift is just over (so let's say the shift took until 13:15, on 13:16 you can't access the shift anymore via the homepage). This should be changed to the button showing "Start new shift" if there are no shifts for today, if there are, always show the button "Go to last shift" and redirect to the last (possibly already closed) shift.

KiOui:
Altered currently_active_shift_for_venue to support joining already ended shifts.

Closes #7

### DIFF
--- a/website/orders/templates/orders/start_shift.html
+++ b/website/orders/templates/orders/start_shift.html
@@ -2,9 +2,13 @@
 {% if venues %}
     {% for next_venue in venues %}
         {% with next_venue.venue|currently_active_shift_for_venue as venue_shift %}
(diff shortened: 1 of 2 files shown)


JobDoesburg:
@KiOui don't you want to change the message on the button as well? To something like 'check out previous shift admin'

KiOui:
Good point, will do that before merging.

PR #19, opened 2020-05-05
Deletion/remove image field

### LINKED ISSUE #12: Remove image field from venue
It turns out we won't be using these images in this UI

KiOui:
Deleted the image field for venues and added migrations.

Closes #12


website/orders/migrations/0002_auto_20200505_1243.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ("orders", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="shift",
+            name="end_date",
+            field=models.DateTimeField(default=datetime.datetime(2020, 5, 5, 13, 15)),
```
JobDoesburg, line 17: Why is this the default value? It appears to be a Django bug maybe with the auto_now setting for these fields?

KiOui: Hmm weird, will take a look

PR #43, opened 2020-05-08
Added a join shift view

### LINKED ISSUE #15: Automatically set assignees if starting/joining shift
Whenever a baker starts or joins a shift, that user should be added to the assignees of the shift as well. 

Maybe we also need a confirmation for this, to prevent people hitting the 'join shift' by accident, making them assigned

KiOui:
Added a view to ask a baker clicking "Join shift at ..." if he/she really wants to join the shift as a baker or just wants to see the admin view.

Closes #15


website/orders/views.py
```diff
@@ -299,6 +303,55 @@ def get(self, request, **kwargs):
...
+
+        confirm = request.POST.get("confirm", None)
+        if confirm == "Yes":
+            assignees = shift.assignees.all()
+            if request.user not in assignees:
+                shift.assignees.add(request.user)
+                shift.save()
+            return redirect("orders:shift_admin", shift=shift)
+        elif confirm == "No":
+            return redirect("orders:shift_admin", shift=shift)
```
JobDoesburg, line 350: I think non-bakers shouldn't be able to see the shift admin (why would they need to?)

KiOui: All staff people can click "Join shift", maybe people are interested in seeing the backend or clicked it on accident, this helps avoid these situations.

JobDoesburg: Seeing the backend is just the same as seeing the front end. I think they should be redirected back.

PR #48, opened 2020-05-09
Added footer menu

### LINKED ISSUE #13: Admin footer menu
I think it would be a good design choice to have all front-end admin actions in a special admin menu in the footer (see the mock designs).

This means:

- At the home page:
    - The start / join shift buttons
- At the shift overview page
    - A button to join baking the shift
- At the shift admin
    - A button to switch the allow/disallow order status
    - A button to extend the shift end date with 10 minutes
    - A button to increase the shift capacity (if almost reached)
    - A button to place an order yourself
    - Maybe more....?

### LINKED ISSUE #14: get_queue() for shifts
The model should decide on the order of the queue. This for example also involves placing bakers (staff users) on top of the queue, while the rest can be sorted by order date

### LINKED ISSUE #26: Footer stick to window bottom without covering page content
The footer of the page should stick to the bottom of the browser window. However, it currently can cover page content. That should not happen and requires some CSS

KiOui:
Added the following:

- A footer menu with buttons for the admin and shift overview page
- Altered the footer and shift header such that they can both be included with a template tag and they will automatically refresh every 5 seconds (javascript is included)
- Altered all javascript such that page building happens server side instead of client side
- Altered the shift overview css such that the css is cleaner and the buttons/status button expand for smaller screen sizes

Closes #13 
Closes #26 
Closes #14


website/orders/models.py
```diff
@@ -178,6 +178,15 @@ class Shift(models.Model):
...
     assignees = models.ManyToManyField(User)
 
+    @property
+    def orders(self):
+        """
+        Get the orders of this shift.
+
+        :return: a Queryset with the orders of this shift.
+        """
+        return Order.objects.filter(shift=self).order_by("user", "created")
```
JobDoesburg, line 188: You want to put staff users on top! (and then you also close #14 )

website/orders/templates/orders/shift_overview.html
```diff
@@ -1,42 +1,42 @@
...
+{% endblock %}
+
+{% block footer %}
+    {% if user.is_staff and user not in shift.get_assignees %}
+        <footer class="page-footer navbar navbar-expand-md fixed-bottom">
+            <div class="container text-center">
+                <div class="row flex-grow-1">
+                    <div class="col">
+                        <div class="btn-ml m-auto cursor-pointer btn-on" onclick="window.location.href = '{% url 'orders:shift_join' shift=shift %}';">
+                            <i class="fas fa-sign-in-alt"></i>
```
JobDoesburg, line 21: I still suggest to change some of these icons to the alternatives I sent you via whatsapp

JobDoesburg: They are still a bit unclear now, and imo a UI should never be explained

website/tosti/templates/tosti/footer_buttons.html
```diff
@@ -0,0 +1,17 @@
+<footer class="page-footer navbar navbar-expand-md fixed-bottom">
+    <div class="container text-center">
+        <div class="row flex-grow-1">
+            {% for button in buttons %}
+                <div class="col">
+                    <div class="btn-ml btn-on m-auto">
+                        {% if button.icon %}
+                            <i class="fas fa-{{ button.icon }}"></i>
```
JobDoesburg, line 8: Okay that comment should have been here actually

JobDoesburg: Nice code though

JobDoesburg:
Suggestion: maybe make the footer menu collapsable? Then we can allow ourselves to make bigger buttons. And it is better UX since as a baker, you most often don't need to interact with the admin buttons only with the queue

website/tosti/templates/tosti/footer_buttons.html
KiOui: This file should have been deleted :laughingstocks

KiOui:
The footer menu is now collapsable

JobDoesburg:
Maybe center the collapse icon horizontally

JobDoesburg:
Also, the success messages now do not go away. So if I add 5 minutes, it stays there, though I think the message should disappear after some seconds

PR #49, opened 2020-05-10
Added a remember username switch

### LINKED ISSUE #34: Remember my username
It feels very stupid to keep entering your science username when you want to login. Better: give people the option to store their science username in a cookie. Whenever you want to login, it will already be prefilled, allowing even faster login (which is why we want OpenID)

KiOui:
Added a checkbox on the login page to remember a username.

Closes #34


website/users/views.py
```diff
@@ -15,6 +15,8 @@ class LoginView(TemplateView):
 
     template_name = "users/login.html"
 
+    remember_cookie = "_remember_username"
```
JobDoesburg, line 18:
```suggestion
    remember_cookie = "_remembered_username"
```

PR #50, opened 2020-05-10
Added drop-out-of-header logo styling

### LINKED ISSUE #9: Increase site logo size
It would be really nice to follow the mock design for the navbar, where the logo is larger than the navbar and falls out of the navbar on the bottom. However this requires some harder CSS because of alignment and such (the header size is normally inherited from its children).

### LINKED ISSUE #45: Drop-out-of-menu-bar logo
It would be nice if the logo would be positioned such that it would drop out of the menu bar a little bit.

KiOui:
Added CSS such that the logo "hangs" out of the menu bar.

Closes #45 
Closes #9


JobDoesburg:
image

I think it needs just a little more (so a bit more padding at the top of the page content

JobDoesburg:
Maybe we should now also add drop shadow to the menu bar

JobDoesburg:
By the way, does clicking the logo do anything now? Maybe it would be fun to add some javascript to let the logo spin 360 degrees or something (does not need to be in this PR obviously)

KiOui:
> By the way, does clicking the logo do anything now? Maybe it would be fun to add some javascript to let the logo spin 360 degrees or something (does not need to be in this PR obviously)

Nice idea!

PR #63, opened 2020-05-13
Added a staff required permission

### LINKED ISSUE #56: Superuser status is needed for anything else then ordering items
Someone needs superuser status for joining shifts/changing them, etc. A 403 Forbidden is returned.

KiOui:
Added a staff-based permission and edited the permissions of the order application to staff-based permission views.

Closes #56


KiOui:
We could also do it like this, this would actually split up where we define permissions into both views.py and urls.py, I think the first solution I proposed is nicer then this one.

JobDoesburg:
> We could also do it like this, this would actually split up where we define permissions into both views.py and urls.py, I think the first solution I proposed is nicer then this one.

Wouldn't it become like

```python
@permission_required('is_staff')
def some_method(....)
```
?

KiOui:
> > We could also do it like this, this would actually split up where we define permissions into both views.py and urls.py, I think the first solution I proposed is nicer then this one.
> 
> Wouldn't it become like
> 
> ```python
> @permission_required('is_staff')
> def some_method(....)
> ```
> 
> ?

Yes but that does not work for class-based views, so its either the state of the branch right now or one commit earlier with a class-based permission we created ourselves.

JobDoesburg:
> > > We could also do it like this, this would actually split up where we define permissions into both views.py and urls.py, I think the first solution I proposed is nicer then this one.
> > 
> > 
> > Wouldn't it become like
> > ```python
> > @permission_required('is_staff')
> > def some_method(....)
> > ```
> > 
> > 
> > ?
> 
> Yes but that does not work for class-based views, so its either the state of the branch right now or one commit earlier with a class-based permission we created ourselves.

Okay just do what you think is best

KiOui:
Okay, then we'll go with the first option, I reverted the commit.

PR #64, opened 2020-05-15
Changed the order view to something more nice

KiOui:
Schermafbeelding 2020-05-15 om 22 58 31
Changed the order view to something a little bit more "intuitive". The buttons are now styled more nicely, the interface now has two columns on desktop view and the alerts are more specific.


JobDoesburg:
I think the 'back to overview' button should be fully centered

KiOui:
What about a back to shift overview button in the footer? And then a "order items" button in the footer of the shift overview.

JobDoesburg:
> What about a back to shift overview button in the footer? And then a "order items" button in the footer of the shift overview.

Then I would say add some breadcrumbs on top and include the button there. The footer should only be for admins in my opinion (but breadcrumbs might be nice, that always works when page structure is unclear)

KiOui:
Hmm, could you make a mock-up, I'm not sure what you mean.

KiOui:
I'm just going to center the button, we can look at breadcrumbs later.

PR #65, opened 2020-05-17
Added a shift admin button

### LINKED ISSUE #23: Django Admin to frontend
There should be an easy button for staff users to switch from Django Admin to front end view for shifts. So the admin front end view should have a settings icon in the admin menu to switch to the Django Admin for that shift, and the other way around as well!

KiOui:
Added a shift admin button on the frontend shift admin page.

Closes #23 

Schermafbeelding 2020-05-17 om 21 11 33


JobDoesburg:
Nice! I would only rephrase it to something like “go to more detailed settings” with a settings icon

KiOui:
Should I also implement a button on the Django admin side to switch to the frontend admin side? If so, the easiest way is at the bottom but that might be weird as the save buttons are there, what do you think @JobDoesburg?

JobDoesburg:
Yeah, that is also an issue that’s open. The native way would be to place it in the object tools. There is some way to do this nicely in django by specifying some field I know but not sure how. Concrexit has it for events

PR #66, opened 2020-05-20
Changed the starting time to the time it is now

### LINKED ISSUE #46: Start shift view should suggest datetime.now() as start time instead of default values
Start shift view should suggest datetime.now() as start time instead of default values.

Currently, when you hit the 'start shift' button, the time of the shift is overwritten to default values. This is not desirable when creating a shift via the 'start shift' button since you want to start a shift right at that moment. 

Maybe the form should also include a checkbox whether orders are immediately allowed or not

KiOui:
Changes the default starting time of the start shift form to the time it is currently, rounds down to the minute.

Closes #46

### DIFF
--- a/website/orders/forms.py
+++ b/website/orders/forms.py
@@ -1,6 +1,7 @@
 from django import forms
 from django.contrib.auth import get_user_model
 from .models import Shift
+from datetime import datetime, timedelta
 
 
 User = get_user_model()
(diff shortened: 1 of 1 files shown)


JobDoesburg:
What would happen with the end time? Could that now be in the past? That would fail

KiOui:
Oehh very good one, what should we do? Set it to the next occurrence of 13:30?

JobDoesburg:
I think set it to +1 hour…
And then maybe do it differently if the shift is started between 12:00 and 13:30 or something like that?

KiOui:
Or maybe default it back to 12:30 if it is past 13:30

KiOui:
Yeah that is also a good idea

KiOui:
Erg huts van je

PR #69, opened 2020-05-31
Feature/spotify integration

### LINKED ISSUE #28: Feature: integration with Marietje
Why have 2 systems if 1 is sufficient?

### LINKED ISSUE #68: Unified way of refreshing page areas
We are now using a lot of different views and javascript files to refresh parts of the page. We might be able to do this in a more unified way using one request for a refresh of all the page areas that need refreshing.

KiOui:
Added an audio controller for Spotify. Currently featuring:

- A new "Marietje" application
- Integrated Spotify OAuth
- Backend selection of playback device
- Rewrite of javascript such that a lot of files are not needed anymore
- Easy to use inclusion tags for the player and queue attribute

Closes #28 
Closes #68


website/marietje/templates/marietje/index.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'tosti/base.html' %}
+{% load static queue %}
+
+{% block page %}
+    <div class="container-sm">
+        <h1 class="mb-3"><i class="fas fa-music"></i> Marietje</h1>
+        <p class="text-left">Marietje is the music control system for the student canteens in the Huygens building.
```
JobDoesburg, line 7: Maybe this text should be less history and more 'what can you do now'

website/marietje/models.py
```diff
@@ -0,0 +1,179 @@
...
+            return self.client_id
+
+    class Meta:
+        """Meta class."""
+
+        verbose_name = "Spotify settings"
+        verbose_name_plural = "Spotify settings"
+
+
+class SpotifyArtist(models.Model):
```
JobDoesburg, line 133: Is this model really required

website/marietje/templates/marietje/authorize.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'tosti/base.html' %}
```
JobDoesburg, line 1: Do we want this to be a front end page? I think a admin page might be nicer / make more sense

website/marietje/templates/marietje/index.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'tosti/base.html' %}
+{% load static queue %}
+
+{% block page %}
+    <div class="container-sm">
+        <h1 class="mb-3"><i class="fas fa-music"></i> Marietje</h1>
```
JobDoesburg, line 6: Maybe use Spotify icon here too

website/marietje/admin.py
```diff
@@ -0,0 +1,46 @@
...
+    """Spotify Track admin."""
+
+    list_display = ["track_name", "track_id"]
+
+
+@admin.register(SpotifyQueueItem)
+class SpotifyQueueItemAdmin(admin.ModelAdmin):
+    """Spotify Queue item admin."""
+
+    list_display = ["track", "spotify_settings_object", "requested_by", "added"]
```
JobDoesburg, line 46: These admins are not the most intuitive for the user

website/marietje/templates/marietje/index.html
KiOui: Good point, will adjust

website/marietje/models.py
KiOui: It might be nice for some statistics later on

website/marietje/templates/marietje/authorize.html
KiOui: I’m not sure how you would see this as an admin page, also I really like how the page looks like this. Maybe we can adjust the header so that it does not seem like a “normal” page?

website/marietje/templates/marietje/index.html
KiOui: Good point, will change

website/marietje/admin.py
KiOui: Will remove

website/marietje/templates/marietje/authorize.html
KiOui: I took a look at how to do this, this is really difficult so if you agree we can create an issue for this and do this at a later stage

website/marietje/models.py
JobDoesburg: Okidoki

website/marietje/models.py
```diff
@@ -0,0 +1,180 @@
...
+from django.contrib.auth import get_user_model
+from django.db import models
+from django.conf import settings
+from spotipy import SpotifyOAuth
+from spotipy.client import Spotify
+
+User = get_user_model()
+
+
+class SpotifySettings(models.Model):
```
JobDoesburg, line 12: I think `SpotifySettings` is a bit unclear. Maybe call it `SpotifyAccounts` (because that's basically what it is). You add/link Spotify accounts to django and then you can use those at certain venues.

website/marietje/templates/marietje/authorize.html
JobDoesburg: wouldn't it just work if you would extend the admin base template?

JobDoesburg: if not, fine by me

website/marietje/admin.py
```diff
@@ -0,0 +1,25 @@
+from django.contrib import admin
+from django.shortcuts import redirect
+
+from .models import SpotifySettings
+from .forms import SpotifySettingsAdminForm
+
+
+@admin.register(SpotifySettings)
+class SpotifySettingsAdmin(admin.ModelAdmin):
```
JobDoesburg, line 9: Maybe add a venue-picker here as well, would be nicer than the other way around

PR #72, opened 2020-06-16
Added a user details page

### LINKED ISSUE #70: Display usernames differently
Usernames are now displayed as they are known at CNCZ. The bakers of a shift should now the first name of someone that orders a product, so they can shout it through the canteen if their tosti is ready. We should create a way in which people enter there first/last name and change the way it is displayed in the order views.

### LINKED ISSUE #71: Add a to_lowercase on science login name
science usernames are always lowercase, but OpenID is capital-sensitive. So it would be nice to make this conversion in our code

KiOui:
Added a page where users can see their user information.
Changed the OpenID functions and created a simple class.
Changed the OpenID request such that user information is automatically requested and stored. 

Closes #70 
Closes #71


JobDoesburg:
Can't we just request the full name of the user via OpenID?

JobDoesburg:
If not I think we should build some check on the lastname and the user name to have some fraud check. Or at least show names as `Firstname Lastname (username)`, e.g. "John Doe (jdoe)" so the username remains visible and we do have some unique part in all displayed names

KiOui:
> Can't we just request the full name of the user via OpenID?

I don't think this is possible

JobDoesburg:
It is, by adding `"openid.sreg.required": "fullname,email"` to the parameters in the request

JobDoesburg:
or maybe better even `email_verified` if that works, would need to check

website/users/models.py
```diff
@@ -61,7 +61,12 @@ def __str__(self):
 
         :return: the username of the user
         """
-        return self.username
+        if self.first_name and self.last_name:
+            return "{} {}".format(self.first_name, self.last_name)
+        elif self.first_name:
+            return self.first_name
+        else:
+            return self.username
```
JobDoesburg, line 69: Django's built-in get_full_name() works for this right?

website/users/views.py
```diff
@@ -67,6 +61,26 @@ class VerifyView(TemplateView):
 
     template_name = "users/verify.html"
 
+    def set_user_details(self, openid_verifier, user):
```
JobDoesburg, line 64: Let's move this to services

website/users/forms.py
```diff
@@ -17,4 +17,23 @@ def clean_username(self):
...
+        help_text="This is your username used to log you in to this website.",
+        widget=forms.TextInput(attrs={"readonly": "readonly"}),
+    )
+    first_name = forms.CharField(
+        label="First name", widget=forms.TextInput(attrs={"readonly": "readonly"}),
+    )
+    last_name = forms.CharField(
+        label="Last name", widget=forms.TextInput(attrs={"readonly": "readonly"}),
+    )
+    email = forms.CharField(
```
JobDoesburg, line 37: I think there is an email field for this? At least we should add an email validator

website/users/views.py
```diff
@@ -77,16 +91,15 @@ def get(self, request, **kwargs):
...
-            except User.DoesNotExist:
-                user = User.objects.create(username=username)
-            login(request, user)
-            return redirect("index")
+        openid_verifier = get_openid_verifier(request)
+        if openid_verifier.verify_request():
+            username = openid_verifier.extract_username()
+            if username:
+                user, created = User.objects.get_or_create(username=username)
+                if created:
```
JobDoesburg, line 99: Maybe move this to services too

website/users/forms.py
KiOui: The user details can not be changed, it is only for displaying.

PR #80, opened 2020-06-25
Feature/marietje improvements

JobDoesburg:
- Fixed migrations
- Admin and front end have been separated
- Some code improvements
- Some improved docs

Sorry for the Git history


KiOui:
Schermafbeelding 2020-06-26 om 12 29 32

These things must be aligned correctly, also maybe make the divider only as big as the text and buttons are.

PR #81, opened 2020-06-26
Separate logic in user module

JobDoesburg:


KiOui:
These things are half already changed in #72, maybe check that out first and put these changes within that merge request?

KiOui:
This is now included in #72 (and changed a little bit)

PR #87, opened 2020-06-28
Permission model

### LINKED ISSUE #29: Permission model
Instead of using the `is_staff` flag to determine whether someone is a baker, and `is_superuser` to check whether they can see the full admin, there should be a better permission model, with groups per venue and venue/shift specific permissions:

- Only assignees to a shift should be able to access a shift
- Users can only assign themselves to a shift during that shift
- Only super users should be able to change a shift after it is finished
- Only people assigned to a venue can start a shift at that venue
- Maybe more permissions should exist...

JobDoesburg:
Closes #29


JobDoesburg:
@KiOui I wrote this first version, but it has some problems with the user model. Django guardian wants us to use `auth.User` and `auth.Group`. So that needs some rewriting to work properly. Currently our `User` class is not an `auth.User` so that is bad. Also, I have extended `auth.Group` into `user.UserGroup`, which also isn't what we want.

Do you have suggestions how to do it better?

website/tosti/settings/base.py
```diff
@@ -28,14 +28,20 @@
...
     "import_export",
+    "guardian",
     "marietje",
     "users",
     "tosti",
     "venues",
     "orders",
 ]
 
-AUTH_USER_MODEL = "users.User"
```
KiOui, line 38: Changing this line will make the project not use our user model anymore, that is why we can't overwrite the __str__ and short name and why we are not seeing the normal names but the usernames in the template, we might want to change this to use our User model again.

website/users/models.py
```diff
@@ -44,11 +43,9 @@ def create_superuser(self, username, **kwargs):
         return self._create_user(username, **kwargs)
 
 
-class User(AbstractUser):
+class User(BaseUser):
```
KiOui, line 46: Are we actually using this class anymore? Because the settings file does not state a custom User model anymore.

website/users/migrations/0002_auto_20200706_1400.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.0.5 on 2020-07-06 12:00
+
+import django.contrib.auth.models
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
KiOui, line 7: We might want to squash the migrations in this pull request or before the first release.

PR #91, opened 2020-07-06
Added an explainer for TOSTI

### LINKED ISSUE #24: Explainer
People that never ordered with TOSTI before, should have a page that explains how it works. 

1. You login with your science user name
2. You see at which venues you can order
3. You select the shift you want to order at
4. You see the queue
5. You place your orders, they get added to the queue
6. You can maybe delete your orders if you made a mistake
7. You pay for your orders
8. Your orders are ready and people will call your name / the website will notify you
9. You come get your order
10. Bon apetit

KiOui:
Added two explainers for the orders module, one for users and one for administrators.

Closes #24


JobDoesburg:
Maybe we should rewrite the text to something a bit more engaging. The target audience is students. 

I don't mind thinking about that. 


Looks good though

PR #92, opened 2020-07-06
Some styling fixes

### LINKED ISSUE #35: UX oriented buttons and texts
Some buttons and texts right now have a big focus on the underlying model / code:

- "Back to order screen" 
- "Payment needed"
- "Click to go to status page"
- "Order items"

For UX purposes, it would be better to replace those to things such as 

- "Back to overview"
- "Not paid"
- "Check your order"
- "Place order"

These are just examples and might be outdated, it is however important to keep an eye on these things.

KiOui:
Fixed some styling issues due to the font awesome change.

Closes #35


JobDoesburg:
Maybe remove the whole footer for non-admins?

KiOui:
I might also add a new index view

PR #107, opened 2020-07-08
Redirect 403's to login if not logged in

### LINKED ISSUE #106: Order view returns 403 for non logged in users
The order view returns a 403 for non-logged-in users, this should redirect to the login view.

KiOui:
Closes #106


JobDoesburg:
Maybe redirect with a `next` parameter so that after login in, the user is redirected?

PR #124, opened 2020-08-10
Added a barcode scanner

### LINKED ISSUE #119: Fixtures
To simplify development, we should implement fixtures

### LINKED ISSUE #122: Feature: Addition of items that can be bought without limit
Items such as soda cans or chocolate bars can be bought in the canteens unlimited during shifts. We need to add items that have no limits to the item list. Also, shift admins should be able to add items bought in the canteen "generally" (so without a user connected). Also there should be a choice whether or not to display ordered items in the queue of the shift.

This all to make the transition from Thadmin to TOSTI better.

### LINKED ISSUE #123: Feature: Adding a barcode scanner to the shift admin page
For scanning barcodes of soda cans or chocolate bars (etc.).

KiOui:
Added the following:

- A couple of field in the Product class to support Orderable and Scannable products. Orderable products appear on the Order page, Scannable products can be added by scanning them with a barcode scanner.
- The addition of a ignore_shift_restrictions field in the Product class to support products that may ignore shift restrictions
- The addition of a type field in the Order class to support different types of orders (for now only Ordered and Scanned)
- The addition of a barcode scanner on the admin page
- An alteration of the order overview where scanned items appear with a desktop fontawesome icon
- A fixtures file (yes, really nice isn't it 😄)

Closes #119 
Closes #122 
Closes #123


.gitignore
```diff
@@ -148,3 +148,4 @@ docker-compose.yml
 # Documentation
 /docs/build
 
+
```
JobDoesburg, line 151: ?

website/orders/models.py
```diff
@@ -100,7 +100,9 @@ class Product(models.Model):
     name = models.CharField(max_length=50, unique=True, null=False, blank=False)
     icon = models.CharField(
         max_length=20,
-        unique=True,
+        unique=False,
```
JobDoesburg, line 103: If this isn't unique anymore, maybe we should also show the product name in the shift admin then. Because now we only display the customer name and icon. If the icon isn't unique we might get ambiguity

website/orders/models.py
```diff
@@ -110,6 +112,15 @@ class Product(models.Model):
         max_digits=6, decimal_places=2, validators=[MinValueValidator(Decimal("0.00"))]
     )
 
+    orderable = models.BooleanField(
+        default=True, null=False, help_text="Whether or not this product should appear on the order page."
```
JobDoesburg, line 116: Maybe refactor this to 'is_takeaway' or something, as in "a product that needs to be prepared".

website/orders/models.py
```diff
@@ -119,6 +130,22 @@ class Product(models.Model):
         "bound to the venue. Empty means no limit.",
     )
 
+    scannable = models.BooleanField(default=False, help_text="Whether or not this product can be scanned.")
```
JobDoesburg, line 133: Can be removed, as a barcode value of Null implies this

website/orders/models.py
```diff
@@ -119,6 +130,22 @@ class Product(models.Model):
...
+    barcode = models.CharField(null=True, blank=True, max_length=13, help_text="Either an EAN-8 or EAN-13 barcode.")
```
JobDoesburg, line 134: Maybe add a validator for 8 digits or 13 digits. Would a positiveintegerfield work better maybe?

website/orders/models.py
```diff
@@ -119,6 +130,22 @@ class Product(models.Model):
...
+    def save(self, *args, **kwargs):
+        """
+        Save an object of the Product type.
+
+        :param args: arguments
+        :param kwargs: keyword arguments
+        :return: None, raises a ValueError on error
+        """
+        if self.scannable and (self.barcode is None or self.barcode == ""):
+            raise ValueError("A barcode must be specified for products that can be scanned.")
```
JobDoesburg, line 145: Can be removed (see above). Otherwise, this should be in the clean() method, not in save()

website/orders/models.py
```diff
@@ -498,6 +567,8 @@ class Order(models.Model):
     paid = models.BooleanField(default=False, null=False)
     paid_at = models.DateTimeField(null=True, blank=True)
 
+    type = models.PositiveIntegerField(choices=TYPES, default=0, null=False, blank=False)
```
JobDoesburg, line 570: This shouldn't be here as it can be derived from the product (check if the product is orderable/takeaway)

website/orders/static/orders/js/admin-scanner.js
```diff
@@ -0,0 +1,179 @@
...
+    MESSAGE_CONTAINER.innerHTML = "";
+    MESSAGE_CONTAINER.append(create_element("p", ["alert", "alert-success"], "Product added successfully!"))
+}
+
+function product_added_failed(errormsg) {
+    MESSAGE_CONTAINER.innerHTML = "";
+    MESSAGE_CONTAINER.append(create_element("p", ["alert", "alert-danger"], errormsg))
+}
+
+start_scanner();
```
JobDoesburg, line 179: I think it would be nicest to make the scanner page a simple form with a single field for the product code (that can be entered manually!), and with this scanner only filling that field. 
I think that would make the whole thing more user friendly, as you can also enter the code manually if you want.

The 'confirm the product' screen is undesirable I think, as it slows down the flow. What do you think? Maybe only give a 'undo' button?

website/tosti/fixtures/default.json
```diff
@@ -0,0 +1,206 @@
...
+},
+{
+    "model": "users.groupsettings",
+    "pk": 4,
+    "fields": {
+        "is_auto_join_group": true,
+        "gets_staff_permissions": false
+    }
+}
+]
```
JobDoesburg, line 206: Gaaf!

website/orders/models.py
JobDoesburg: Also, let's make this unique!

JobDoesburg:
@KiOui For some reason, when testing on iPhone, it doesn't work correctly anymore (it doesn't ask for permission correctly)

.gitignore
KiOui: I thought an extra space would be nice, will be fixed in the next commit.

website/orders/models.py
KiOui: Hmm, I had some troubles with the icons. What should we give all soda cans for example? A different one each? The admin panel will now show a product name if no icon is provided. I can do two things:

1. Always show the product name (but then we might have to do some CSS trickery for the layout)
2. Make this unique again (but then the soda cans can't all have the same icon)

I prefer 1. What do you think?

website/orders/models.py
KiOui: I thought of it as "can be ordered from the order page". Maybe rename to "can_be_ordered"?

We can namely also let the soda cans be ordered with the tosti's on the order page.

website/orders/models.py
KiOui: True

website/orders/models.py
KiOui: A validator is a good idea, I will also change it to a PositiveIntegerField.

website/orders/models.py
KiOui: Noted

website/orders/models.py
KiOui: Yes but a product can be orderable AND scannable. This is merely the way the product was registered (for example, scanned orders usually have no user associated with them).

website/orders/static/orders/js/admin-scanner.js
KiOui: The problem is that the scanner is not really fault-proof. I'm not sure what the barcodes of the items in the canteen look like but if there are two that really look alike the scanner might pick something up that should not be there and might scan the wrong item. This is why I implemented the "select the product screen".

I can change it to this:

- The scanner automatically starts
- If the scanner picks up a correct barcode, it is added to a list of items at the bottom of the page.
- There are crosses next to the items that let the user delete items that are not wanted
- If a user clicks "Add to shift" at the bottom of the page, all items of the list get added.

Sound good?

website/orders/models.py
JobDoesburg: Let's go for 1. (and maybe if the CSS is too hard, we could add a shortname field to products and only display that)

website/orders/models.py
JobDoesburg: But then they will also appear in the queue, right?

website/orders/models.py
JobDoesburg: I see... We should document this, as it is pretty confusing

(You can now have orders:

- with and without a user attached
- scanned or 'ordered via the form' (I think it is not in order ordering ordered orders)
- products that can be ordered or not

this feels unnecessarily complex)

website/orders/static/orders/js/admin-scanner.js
JobDoesburg: > The problem is that the scanner is not really fault-proof. I'm not sure what the barcodes of the items in the canteen look like but if there are two that really look alike the scanner might pick something up that should not be there and might scan the wrong item. This is why I implemented the "select the product screen".
> 
EAN codes have a checksum so this shouldn't be a problem I think.

Benefit of such a field would be that we can also easily support 'real' manual scanners


> I can change it to this:
> 
> * The scanner automatically starts
> * If the scanner picks up a correct barcode, it is added to a list of items at the bottom of the page.
> * There are crosses next to the items that let the user delete items that are not wanted
> * If a user clicks "Add to shift" at the bottom of the page, all items of the list get added.
> 
> Sound good?

Sounds more complex than what we currently have, so lets not do this then.

website/orders/models.py
KiOui: This might be resolved in the latest commit.

website/orders/models.py
KiOui: Hmm, but it might be nice to order a soda can with your TOSTI on the webpage and also display the list of things that can be bought in the canteens. As of now it will appear on the order page. I suggest we change it to the following:

- We add a "show_on_shift_overview" BooleanField.
- Users are able to order soda cans on the order page.
- The orders of products with show_on_shift_overview False will result in the orders not being shown to all users, but rather only to the user that ordered the product.
- The admin view shows all orders: Tosti's, ordered soda and scanned soda.

website/orders/models.py
KiOui: Yes but don't you think it might be handy in the future to store the way an order was registered. We can also set the user field to the one who scanned the order if the type = Scanned. This will remove some of the options you pointed at.

website/orders/static/orders/js/admin-scanner.js
KiOui: > > The problem is that the scanner is not really fault-proof. I'm not sure what the barcodes of the items in the canteen look like but if there are two that really look alike the scanner might pick something up that should not be there and might scan the wrong item. This is why I implemented the "select the product screen".
> 
> EAN codes have a checksum so this shouldn't be a problem I think.

Ahh, nice, then I'll implement a checksum checker in javascript (which will be a lot of fun, not).

> Benefit of such a field would be that we can also easily support 'real' manual scanners
> 
> > I can change it to this:
> > 
> > * The scanner automatically starts
> > * If the scanner picks up a correct barcode, it is added to a list of items at the bottom of the page.
> > * There are crosses next to the items that let the user delete items that are not wanted
> > * If a user clicks "Add to shift" at the bottom of the page, all items of the list get added.
> > 
> > Sound good?
> 
> Sounds more complex than what we currently have, so lets not do this then.

Okay, another idea:

- The scanner will NOT open a new page but rather a popup.
- The scanner will check the EAN codes for the checksum, if a checksum is correct it will be send to the server to look up the associated item.
- If the barcode exists in the database, we will add it immediately, close the scanner and put a notification on top of the page saying the product was added successfully.
- I will also put a search field on top of the scanner where an admin can search for either the barcode or the product name and add it manually.

KiOui:
> @KiOui For some reason, when testing on iPhone, it doesn't work correctly anymore (it doesn't ask for permission correctly).
> 
> Also on mobile, I think it would be nice to make the camera view 100% width

I'll first fix all issues above and then look at this.

website/orders/static/orders/js/admin-scanner.js
JobDoesburg: Sounds good, maybe even faster to do the matching on-device in javascript already. 
Maybe use a javascript alert()?

JobDoesburg: By the way, I assume that our library does an automatic checksum check already

website/orders/models.py
JobDoesburg: > "display the list of things that can be bought in the canteens" 
I agree with that!

website/orders/static/orders/js/admin-scanner.js
KiOui: > Sounds good, maybe even faster to do the matching on-device in javascript already.
> Maybe use a javascript alert()?

Then we need to ship a list of all objects with barcode to the web browser which is not an issue with a couple of products but if we have a lot it is not the best way of doing this.

Django is fast enough to respond almost instantly so it's not a problem to check serverside.

KiOui: > By the way, I assume that our library does an automatic checksum check already

I thought that too but I saw some weird codes ending up in the logs, I'll make sure to check this.

KiOui: > > By the way, I assume that our library does an automatic checksum check already
> 
> I thought that too but I saw some weird codes ending up in the logs, I'll make sure to check this.

You were correct, it checks the codes.

KiOui: This is also changed in the latest commit, could you please check again if you agree with the changes?

website/orders/admin.py
```diff
@@ -37,6 +38,16 @@ class ProductAdmin(admin.ModelAdmin):
...
 
+    def clean_barcode(self):
+        """Clean barcode."""
+        if self.cleaned_data["barcode"] == "" or self.cleaned_data["barcode"] is None:
+            return None
+
+        if len(self.cleaned_data["barcode"]) == 8 or len(self.cleaned_data["barcode"]) == 13:
+            return self.cleaned_data["barcode"]
+        else:
+            raise ValidationError("A barcode must be either 8 or 13 integers long")
```
JobDoesburg, line 49: I think a validator ([https://docs.djangoproject.com/en/3.0/ref/validators/)(https://docs.djangoproject.com/en/3.0/ref/validators/)) would be better as it is a restriction on the model-level and not admin-level

website/orders/models.py
JobDoesburg: I agree

website/orders/admin.py
KiOui: Will change.

KiOui: Okay, had to make it a CharField again as we can't get the len(int) of an integer. A char field is also more nice in the admin panel for a barcode.

KiOui: Done, it also validates the barcode checksum now.

website/orders/models.py
KiOui: And what about this?

> * We add a "show_on_shift_overview" BooleanField.
> * Users are able to order soda cans on the order page.
> * The orders of products with show_on_shift_overview False will result in the orders not being shown to all users, but rather only to the user that ordered the product.
> * The admin view shows all orders: Tosti's, ordered soda and scanned soda.

website/orders/validators.py
```diff
@@ -0,0 +1,33 @@
...
+
+    counter = 0
+    for index, digit in enumerate(value[: len(value) - 1]):
+        if index % 2 == 0:
+            counter += int(digit) * 3
+        else:
+            counter += int(digit)
+
+    if (10 - (counter % 10)) % 10 != int(value[len(value) - 1]):
+        raise ValidationError("The checksum of the barcode is not correct")
```
JobDoesburg, line 33: Maybe a new file is a bit overkill?

JobDoesburg:
- [ ] When scanning an item, the price should appear too (maybe for all products have the `__str__` be `f"{self.name} ({self.price})"`)
- [ ] the close button of the scanner window looks weird on mobile, like its italic

KiOui:
> @KiOui For some reason, when testing on iPhone, it doesn't work correctly anymore (it doesn't ask for permission correctly).
> 
> Also on mobile, I think it would be nice to make the camera view 100% width

This was because the site was not secure, non-https sites will not be allowed to stream from the camera.

PR #132, opened 2020-11-26
Added association app to TOSTI

### LINKED ISSUE #129: Store association of users
Maybe for statistics purposes (and future features) it would be nice to store the association of TOSTI users. On their first login, they should be redirected to their profile page and asked to set their association (not required).

Implementation:

A new `associations` app, and a new Field in the user model.

KiOui:
Added the following:

- A Profile model to store profile information about a User (for example the Association they are in)
- A Association application and model to store association information with the User
- A Welcome screen for new users

Closes #129


JobDoesburg:
I think we should add this to the privacy policy and data minimization command as well. And maybe display the association name in the user's `__str__` (so like "Job Doesburg (Thalia)"). 

Also, as people can change their association, maybe store a copy of the association in an order when we create it (for the statistics).

JobDoesburg (approved):

PR #133, opened 2020-12-11
Restful API for orders

### LINKED ISSUE #101: Refactor: refresh views should return proper JSON, not HTML-in-JSON
Even though it is a nice feature, it is bad coding style in my opinion. The JSON API should deliver proper JSON

### LINKED ISSUE #125: Enhancement: More structured POST views
The POST views that handle changing shift parameters via javascript calls (AJAX) are now mixed up and not intuitive. We need to refactor all views that handle AJAX calls and make them function the same.

This might also directly close #101

KiOui:
Added a Restful API for orders application.

Added the following:

- API views for listing, creating, retrieving and updating Shift objects
- API views for listing, creating, retrieving, updating and destroying Order objects
- API views for listing Product objects
- API views for the features available on the Shift admin screen
- Serializers for all the above methods
- VueJS to easily manage the state of the HTML document in the browser
- A CustomAutoSchema to make adjustments to the provided AutoSchema
- SwaggerUI

Removed the following:

- All views that we were using to provide API-like features that are now implemented in the API
- A couple of templates that are now included in bigger ones to optimize the VueJS and API implementation

To test you can visit `/api/docs`.

Closes #125 
Closes #101


website/orders/api/serializers/orders.py
```diff
@@ -0,0 +1,131 @@
...
+        :param validated_data: validated serialized data
+        :return: an Order if the Order could be successfully added, a ValidationError otherwise
+        """
+        try:
+            return add_order(
+                validated_data["product"],
+                validated_data["shift"],
+                validated_data["type"],
+                user=validated_data["user"],
+                paid=validated_data["paid"],
```
JobDoesburg, line 45: shouldn't the API always create unpaid and unready orders, which an admin can then make paid and ready with a separate PATCH?

website/orders/api/serializers/orders.py
```diff
@@ -0,0 +1,131 @@
...
+            "paid_at",
+            "type",
+        ]
+        read_only_fields = ["pk", "created", "user", "order_price", "ready_at", "paid_at"]
+
+
+class ShiftSerializer(serializers.ModelSerializer):
+    """Serializer for Shift objects."""
+
+    def validate(self, data):
```
JobDoesburg, line 76: is this logic required here? can't we use the clean-method from the model?

website/tosti/settings/settings.py
```diff
@@ -38,6 +38,7 @@
     "admin_auto_filters",
     "import_export",
     "guardian",
+    'rest_framework',
```
JobDoesburg, line 41: Can we add swaggerui?

JobDoesburg (changes requested):
How does authentication work?

website/tosti/settings/settings.py
se-bastiaan: This sounds like a marvelous idea

JobDoesburg: @KiOui bump

website/orders/api/v1/views.py
```diff
@@ -0,0 +1,351 @@
...
+        API endpoint for creating a Shift.
+        Permission required: can_manage_shift_in_venue
+        :param request: the request
+        :param args: arguments
+        :param kwargs: keyword arguments
+        :return: super method or PermissionDenied when 'venue' POST data is None or when user does not have permission
+        in indicated venue
+        """
+        venue = request.data.get("venue")
+        if request.user.has_perm("orders.can_manage_shift_in_venue", venue):
```
JobDoesburg, line 214: don't we have a nice permission class for this?

website/orders/api/v1/views.py
```diff
@@ -0,0 +1,351 @@
...
+
+    serializer_class = ProductSerializer
+    queryset = Product.objects.all()
+
+    def get_queryset(self):
+        """Get the queryset."""
+        return self.queryset.filter(available_at=self.kwargs.get("shift").venue)
+
+
+@api_view(["POST"])
```
JobDoesburg, line 256: I think it would be nicer to make this a PUT

website/orders/services.py
```diff
@@ -97,8 +99,8 @@ def add_order(product, shift, order_type, user=None, set_done=False, force=False
             type=order_type,
             user=user,
             user_association=user.profile.association,
-            paid=set_done,
-            ready=set_done,
+            paid=paid,
```
JobDoesburg, line 102: Again, do we need to set 'paid' and 'ready'?

website/orders/api/v1/urls.py
```diff
@@ -0,0 +1,36 @@
...
+
+urlpatterns = [
+    path("shifts", ShiftListCreateAPIView.as_view(), name="shifts_listcreate"),
+    path("shifts/<int:pk>", ShiftRetrieveUpdateAPIView.as_view(), name="shift_retrieveupdate"),
+    path("shifts/<shift:shift>/orders", OrderListAPIView.as_view(), name="orders_list"),
+    path(
+        "shifts/<shift:shift>/orders/<int:pk>",
+        OrderRetrieveUpdateDestroyAPIView.as_view(),
+        name="orders_retrieveupdatedestroy",
+    ),
```
JobDoesburg, line 28: I think it would be nice to add an endpoint /shift/<shift:shift>/orders/<int:pk>/ready to which you can PUT to make the order ready, and idem with paid.

website/orders/api/v1/urls.py
```diff
@@ -0,0 +1,36 @@
...
+    path("shifts/<shift:shift>/orders/add", OrderCreateAPIView.as_view(), name="orders_create"),
+    path("shifts/<shift:shift>/add-time", shift_add_time, name="shifts_add_time"),
+    path("shifts/<shift:shift>/add-capacity", shift_add_capacity, name="shifts_add_capacity"),
+    path("shifts/<shift:shift>/scanner", shift_scanner, name="shifts_scanner"),
+    path("shifts/<shift:shift>/cart-order", OrderView.as_view(), name="shifts_cart_order"),
```
JobDoesburg, line 33: Do we really need this separately? Can't we just leave the whole cart concept to the user/browser and allow batch requests to /orders/add?

JobDoesburg (changes requested):

website/orders/api/v1/urls.py
```diff
@@ -0,0 +1,36 @@
...
+urlpatterns = [
+    path("shifts", ShiftListCreateAPIView.as_view(), name="shifts_listcreate"),
+    path("shifts/<int:pk>", ShiftRetrieveUpdateAPIView.as_view(), name="shift_retrieveupdate"),
+    path("shifts/<shift:shift>/orders", OrderListAPIView.as_view(), name="orders_list"),
+    path(
+        "shifts/<shift:shift>/orders/<int:pk>",
+        OrderRetrieveUpdateDestroyAPIView.as_view(),
+        name="orders_retrieveupdatedestroy",
+    ),
+    path("shifts/<shift:shift>/orders/add", OrderCreateAPIView.as_view(), name="orders_create"),
```
se-bastiaan, line 29: If you put the OrderCreateAPIView and the OrderListAPIView together you can properly implement the creation of orders as a POST request. Decide what you do based on the HTTP method and not the route.

website/orders/api/v1/urls.py
KiOui: This can be done with a PATCH to the endpoint for a specific Order object.

website/orders/api/v1/urls.py
KiOui: We can do that but that makes the Javascript ordering a bit more complicated, and this works as it is right now, so let's just use what we have right now.

website/orders/api/v1/urls.py
KiOui: I did it this way because the OrderRetrieveUpdateDestroy view and the OrderCreate view use different serializers because of the way we create Order objects in the backend. It seemed more natural to create a different API endpoint for both views as they do not rely on the same set of data that they let the API user access or create.

website/orders/api/v1/views.py
KiOui: Don't think so.

website/orders/api/v1/serializers.py
```diff
@@ -0,0 +1,224 @@
...
+            "ready_at",
+            "paid",
+            "paid_at",
+            "type",
+        ]
+        read_only_fields = ["id", "created", "user", "product", "order_price", "ready_at", "paid_at", "type"]
+
+
+class UserSerializer(serializers.RelatedField):
+    """User serializer."""
```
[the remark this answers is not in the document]
KiOui: Done

website/orders/api/v1/serializers.py
```diff
@@ -0,0 +1,224 @@
...
+        model = User
+        fields = [
+            "id",
+            "username",
+        ]
+        read_only_fields = ["id", "username"]
+
+
+class OrderCreateSerializer(serializers.ModelSerializer):
+    """Serializer for creation of Order."""
```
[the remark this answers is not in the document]
KiOui: The OrderCreateAPIView and the OrderRetrieveUpdateDestroyAPIView rely on different pieces of data. For example, the create view only needs a user, product, and type of Order whereas the retrieve of an Order shows the "product" as an "icon" and a "name" attribute. When creating an Order the "product" must be supplied as an "id" only. This difference seemed a bit off and I thus decided to split them up.

website/orders/api/v1/serializers.py
```diff
@@ -0,0 +1,224 @@
...
+
+        model = models.Product
+        fields = [
+            "name",
+            "icon",
+        ]
+        read_only_fields = ["name", "icon"]
+
+
+class UserRelatedField(serializers.ModelSerializer):
```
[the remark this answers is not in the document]
KiOui: Moved to the users application

website/orders/api/v1/permissions.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.permissions import BasePermission
+
+
+class HasGuardianPermission(BasePermission):
```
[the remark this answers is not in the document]
KiOui: Refactored

website/orders/services.py
KiOui: I reverted it to setting paid and ready later.

website/orders/api/v1/views.py
KiOui: Nope

website/orders/api/v1/views.py
KiOui: I altered it to PATCH

website/tosti/settings/settings.py
KiOui: Done

JobDoesburg:
- [ ] We still need to style swagger UI with some TOSTI styling
- [ ] In my opinion the API should accept POST to `/orders` instead of POST to `/cart-order/` or `/orders/add`
- [ ] We should add / improve documentation to the schema

KiOui:
> * [ ]  In my opinion the API should accept POST to `/orders` instead of POST to `/cart-order` or `/orders/add`

The `/cart-order` endpoint will still be necessary, I'll try to add the POST to the `/orders` endpoint.

KiOui:
We also need to implement token support for the authenticator

website/orders/api/v1/views.py
```diff
@@ -0,0 +1,393 @@
...
+    def post(self, request, **kwargs):
+        """
+        Create multiple Orders in one go.
+
+        API endpoint for creating multiple orders in one go (handling of a cart).
+        A "cart" POST parameter must be specified including the ID's of the Products the user wants to order.
+        Permission required: can_order_in_venue
+        :param request: the request
+        :param kwargs: keyword arguments
+        :return: a 201 response code if all orders in the cart were successfully processed. Raises a ValidationError
```
JobDoesburg, line 69: We have pydoc `:raises:`

website/orders/services.py
```diff
@@ -86,7 +88,7 @@ def add_order(product, shift, order_type, user=None, set_done=False, force=False
             pass
         # Shift checks
         if not shift.can_order:
-            raise OrderException("User can not order products for this shift")
+            raise OrderException("This shift is closed.")
         if not shift.is_active:
             raise OrderException("This shift is not active")
```
JobDoesburg, line 93: Isn't this checked by the model?

website/tosti/api/openapi.py
```diff
@@ -0,0 +1,48 @@
+from oauth2_provider.scopes import get_scopes_backend
+from rest_framework.reverse import reverse
+from rest_framework.schemas.openapi import SchemaGenerator, AutoSchema
+
+
+class OAuthSchemaGenerator(SchemaGenerator):
```
JobDoesburg, line 6: Shouldn't this be called the OpenAPISchemaGenerator instead of OAuth (as it is openapi schema)

website/marietje/api/v1/views.py
```diff
@@ -0,0 +1,190 @@
...
+def player_play(request, **kwargs):
+    """
+    Player Play API View.
+
+    Permission required: marietje.can_request
+
+    Start playback on a Player.
+    """
+    player = kwargs.get("player")
+    if request.user.has_perm("marietje.can_request", player):
```
JobDoesburg, line 123: shouldn't this be 'can manage'?

website/marietje/api/v1/views.py
```diff
@@ -0,0 +1,190 @@
...
+    serializer_class = QueueItemSerializer
+    queryset = SpotifyQueueItem.objects.all()
+    pagination_class = StandardResultsSetPagination
+
+    def get_queryset(self):
+        """Get the queryset."""
+        return self.queryset.filter(player=self.kwargs.get("player"))
+
+
+@api_view(["GET"])
```
JobDoesburg, line 59: If we could rewrite these to class-based views, you could add the parameters like the search query to the openapi schema so swagger would understand it (by setting `schema=` for the view)

KiOui:
As stated [here](https://github.com/encode/django-rest-framework/commit/7915485c0d94c1ca9ebe6606b78465fffe45acac), there is no easy way to implement manual schema adjustments to AutoSchema of OpenAPI for now, so we have to write our own.

website/marietje/api/v1/views.py
KiOui: can_control actually but nice catch!

website/marietje/api/v1/views.py
KiOui: Done, only the Player schema is not displayed correctly but we might fix that later.

website/orders/api/v1/views.py
KiOui: I have removed most of them as they were displayed on the SwaggerUI page (and that was ugly).

website/orders/api/v1/serializers.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class OrderSerializer(serializers.ModelSerializer):
+    """Serializer for Order model."""
+
+    product_id = serializers.PrimaryKeyRelatedField(queryset=Product.objects.all(), many=False, read_only=False)
+    product = ProductSerializer(many=False, read_only=True)
+    user = UserRelatedField(many=False, read_only=True)
+
+    def validate_product_id(self, value):
```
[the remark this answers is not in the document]
KiOui: What do you mean?

website/orders/services.py
KiOui: We have to rewrite this thing anyway, as it does not work completely (see the TODO somewhere above this line) so let's do this in a different pull request.

website/marietje/api/v1/views.py
```diff
@@ -0,0 +1,218 @@
...
+        Use this endpoint to search for a Spotify Track. Tracks can be searched via their Spotify id.
+        """
+        player = kwargs.get("player")
+        query = request.GET.get("query", "")
+        request_id = request.GET.get("id", None)
+        try:
+            maximum = int(request.GET.get("maximum", 5))
+        except ValueError:
+            maximum = 5
+        if request.user.has_perm("marietje.can_request", player):
```
JobDoesburg, line 101: APIView has a dedicated `check_permissions` / `check_object_permissions` method for this which I think we should use (see http://www.cdrf.co/3.9/rest_framework.views/APIView.html)

website/marietje/api/v1/views.py
```diff
@@ -0,0 +1,218 @@
...
+                    return Response(status=status.HTTP_503_SERVICE_UNAVAILABLE)
+                return Response(status=status.HTTP_200_OK)
+            else:
+                raise ValidationError("A track id is required.")
+        else:
+            raise PermissionDenied
+
+
+@api_view(["PATCH"])
+def player_play(request, **kwargs):
```
JobDoesburg, line 142: For consistency I think we should also turn this into class based apiviews

JobDoesburg (changes requested):

JobDoesburg (approved):
Ja opzicht nice, we moeten alleen ff een issue openen om wat uitgebreidere documentatie hier toe te voegen en OAuth scopes toe te voegen.

PR #140, opened 2021-02-17
Replaced POST with data

### LINKED ISSUE #139: Adding tracks is broken (at least on swagger UI)
The Marietje add song to queue API endpoint seems broken. Also it needs custom schema data

KiOui:
Fixed an error where we consulted the POST attribute instead of data.

Closes #139


JobDoesburg (commented):
I think the CustomAutoSchema does not work properly, could you verify it? 
image
The track id should show as a parameter and there should be example responses as well

KiOui:
This is a POST parameter and therefor it should be specified within the POST parameters (and thus the request body). The parameters that are specified under "Parameters" in SwaggerUI can only be `query`, `header`, `path` and `cookie` parameters (and thus not POST parameters). See [here](https://swagger.io/specification/) for more details (under Parameter object).

JobDoesburg (approved):

PR #145, opened 2021-02-18
Feature/tantalus client

### LINKED ISSUE #136: Add connection to administration system
The description of this issue will be updated when I have some more information about the administration system we are supposed to add a synchronization to.

KiOui:
Added a `tantalus` model that synchronizes Orders with a [Tantalus](https://github.com/thijsmie/tantalus) instance.

The following changes are in this pull request:

- This pull request adds a `finalized` property to `Shifts`. This property can be set to `True` once, after this property is set to `True` the `Shift` can not be changed in any way again.
- This pull request changes the way that cleaning a `Shift` works by adding a `_clean` method that runs on either `clean` or `save` for `Shifts`.
- This pull request harmonizes `Order` creation by adding/altering three functions in `services.py` in the `orders` application.
- This pull request adds a `tantalus` model for which `TantalusProducts` can be registered (connecting TOSTI `Products` to Tantalus products).
- This pull request also adds a receiver for the `Shift` `post_save` signal which checks if `finalized` is set to `True` and afterwards synchronizes all ordered `Products` with the configured Tantalus instance.

Note that while testing, the Tantalus instance should be started on port 8080 (and not the default 8000 where Tantalus starts on).

Closes #136


KiOui:
@thijsmie Could you maybe also check this pull request and especially the synchronization function in `tantalus/signals.py`?

thijsmie:
Hi @KiOui! This looks great, I will deploy this locally and test it, hopefully over the weekend, else next week.

I have one comment to start with. The Venue in TOSTI should map to a Tantalus POS Endpoint since they represent the same thing. Traditionally sales from South and North canteens are also separate in the bookkeeping so it would be nice to keep that.

Tantalus also logs what user enters a sale, purely to track which sales belong to a potentially breached account. This functionality becomes obsolete as TOSTI will take over that functionality.

KiOui:
> I have one comment to start with. The Venue in TOSTI should map to a Tantalus POS Endpoint since they represent the same thing. Traditionally sales from South and North canteens are also separate in the bookkeeping so it would be nice to keep that.

Done! I have created `TantalusOrderVenue`s to register endpoints per venue.

KiOui:
We need to also update the baker explainer.

README.md
```diff
@@ -99,4 +99,15 @@ Setting a venue is easy, just select the venue from the dropdown list and it wil
 
 3. Using the playback functions
 
-If everything went well, a Spotify player is displayed on the `/Marietje/index` page. Administrators are able to play/pause and skip or revert songs, normal users are not. If you open the Spotify player, a search view is displayed and the recent items added to the queue. The search view can be used to search tracks on Spotify. The queue displays only tracks added via Marietje, as there is currently no way to request the queue from a Spotify playback device.
\ No newline at end of file
+If everything went well, a Spotify player is displayed on the `/Marietje/index` page. Administrators are able to play/pause and skip or revert songs, normal users are not. If you open the Spotify player, a search view is displayed and the recent items added to the queue. The search view can be used to search tracks on Spotify. The queue displays only tracks added via Marietje, as there is currently no way to request the queue from a Spotify playback device.
```
JobDoesburg, line 102: Merge conflict

website/orders/api/v1/views.py
```diff
@@ -269,15 +278,19 @@ def patch(self, request, **kwargs):
         """
         shift = kwargs.get("shift")
         time_minutes = request.data.get("minutes", 5)
-        increase_shift_time(shift, time_minutes)
-        return Response(status=status.HTTP_200_OK, data=ShiftSerializer(shift, context={"request": request}).data)
+        try:
+            increase_shift_time(shift, time_minutes)
+            return Response(status=status.HTTP_200_OK, data=ShiftSerializer(shift, context={"request": request}).data)
+        except ValueError as e:
+            return Response(status=status.HTTP_403_FORBIDDEN, data={"detail": e.__str__()})
```
JobDoesburg, line 285: Why is this changed?

website/orders/api/v1/views.py
```diff
@@ -302,6 +315,37 @@ def patch(self, request, **kwargs):
         return Response(status=status.HTTP_200_OK, data=ShiftSerializer(shift, context={"request": request}).data)
 
 
+class ShiftFinalizeAPIView(APIView):
+    """Shift Finalize API View."""
+
+    schema = CustomAutoSchema(response_schema={"$ref": "#/components/schemas/Shift"})
+    permission_required = "orders.can_manage_shift_in_venue"
+    permission_classes = [HasPermissionOnObject]
```
JobDoesburg, line 323: And IsOnBakerList right #147

website/orders/api/v1/views.py
KiOui: Increasing the shift time can actually now throw an exception (due to shifts being finalized). Therefor we need to check it in the view and return a 403 if the shift can't be changed. I now see that this is the wrong exception caught though (it should be a ValidationError).

README.md
KiOui: Hihi will fix

website/orders/api/v1/views.py
KiOui: Yes, let's fix this after merging #147 or let's fix it in #147 after merging this pull request.

website/orders/api/v1/views.py
KiOui: This is now a ValidationError

website/orders/api/v1/views.py
JobDoesburg: #147 is now approved so do your thing

KiOui: Already on it!

website/orders/api/v1/views.py
```diff
@@ -76,9 +77,9 @@ def post(self, request, **kwargs):
         except ValueError:
             return Response(status=status.HTTP_400_BAD_REQUEST)
         try:
-            place_orders(cart.get_item_list(), request.user, shift)
+            add_user_orders(cart.get_item_list(), shift, request.user)
         except OrderException as e:
-            raise ValidationError(e.__str__())
```
JobDoesburg, line 81: I think we should actually raise an exception, but maybe a `PermissionDenied`. DRF will catch them right?

website/orders/api/v1/views.py
```diff
@@ -134,6 +135,13 @@ def perform_create(self, serializer):
                 shift=shift, order_type=Order.TYPE_ORDERED, user=self.request.user, paid=False, ready=False
             )
 
+    def create(self, request, *args, **kwargs):
+        """Catch the OrderException that might be thrown by creating a new Order."""
+        try:
+            return super(OrderListCreateAPIView, self).create(request, *args, **kwargs)
+        except OrderException as e:
+            return Response(status=status.HTTP_403_FORBIDDEN, data={"detail": e.__str__()})
```
JobDoesburg, line 143: idem

website/orders/api/v1/views.py
JobDoesburg: Idem, I think this can raise an exception

website/orders/api/v1/views.py
```diff
@@ -315,8 +328,46 @@ def patch(self, request, **kwargs):
...
+        except DjangoValidationError as e:
+            return Response(status=status.HTTP_403_FORBIDDEN, data={"detail": e.__str__()})
+
+
+class ShiftFinalizeAPIView(APIView):
+    """Shift Finalize API View."""
+
+    schema = CustomAutoSchema(response_schema={"$ref": "#/components/schemas/Shift"})
+    permission_required = "orders.can_manage_shift_in_venue"
+    permission_classes = [HasPermissionOnObject, IsOnBakersList]
```
JobDoesburg, line 344: After #162 is merged, we should add the same scoping here

website/orders/models.py
```diff
@@ -265,6 +265,14 @@ class Shift(models.Model):
         "possible, even in the given time frame.",
     )
 
+    finalized = models.BooleanField(
```
JobDoesburg, line 268: We should also prevent changing `Order`s themselves when their shift has changed

website/tantalus/signals.py
```diff
@@ -0,0 +1,50 @@
...
+
+from orders.models import Shift, Order
+from tantalus.models import TantalusProduct, TantalusOrderVenue
+from tantalus.services import sort_orders_by_product, get_tantalus_client, TantalusException
+
+
+@receiver(post_save, sender=Shift)
+def sync_to_tantalus(sender, instance, **kwargs):
+    """Synchronize Orders to Tantalus if a Shift is made finalized."""
+    if instance.finalized:
```
JobDoesburg, line 14: This will run every time you save a shift. So it wil run twice if you are saving a shift that is finalized. Instead, you want to check the old instance here

website/tantalus/signals.py
```diff
@@ -0,0 +1,50 @@
...
+        try:
```
JobDoesburg, line 15: I think it would be good to move this whole sync code to a separate service. This way, you can for example run the same code from a management command and it is good for testing

JobDoesburg (changes requested):

website/orders/api/v1/views.py
KiOui: Good point, I'll change them all (it will work the same as DRF will just return a 403).

website/orders/models.py
```diff
@@ -620,50 +661,16 @@ def __str__(self):
...
-                {"user": "A user must be specified if the order type is {}".format(self.TYPES[self.TYPE_ORDERED][1])}
-            )
-
-        if errors:
-            raise ValidationError(errors)
+        Check if the Shift is already finalized.
+        :return:
+        :rtype:
+        """
+        super().clean()
```
JobDoesburg, line 671: I don't see how all those cleans you removed have moved to a super class?

website/tantalus/signals.py
```diff
@@ -0,0 +1,14 @@
...
+from tantalus.models import TantalusShiftSynchronization
+from tantalus.services import synchronize_to_tantalus
+
+
+@receiver(post_save, sender=Shift)
+def sync_to_tantalus(sender, instance, **kwargs):
+    """Synchronize Orders to Tantalus if a Shift is made finalized."""
+    if instance.finalized and not TantalusShiftSynchronization.objects.get(shift=instance).exists():
+        if synchronize_to_tantalus(instance):
+            TantalusShiftSynchronization.objects.create(shift=instance)
```
JobDoesburg, line 14: I think creating the TantalusShiftSynchronization should also move to synchronize_to_tantalus

JobDoesburg (changes requested):

website/tantalus/signals.py
KiOui: I thought about this but it might also be needed to do a synchronize without creating an instance of that class. We can for example add the creation of an instance when creating a management command.

website/orders/models.py
KiOui: This is just good coding practice right? Calling the super method?

website/tantalus/signals.py
JobDoesburg: Sure

website/orders/models.py
JobDoesburg: But you're removing all other checks, like the check if you can still order because of time restrictions, max_orders_per_user restrictions, etc

PR #147, opened 2021-02-20
Added permissions to Shift admin views if user is not in the assignees

### LINKED ISSUE #98: Shift admin view does not throw 403 if user not in assignees
This is unexpected

KiOui:
Shift admin actions and views now require the user to be in the assignees of the shift before working. The following has changed:

- The `orders` API now has a new permission: `IsOnBakersList` which checks if the user is on the assignees list of the `Shift`.
- The following API views are now restricted by this permission: 

- `OrderListCreateAPIView` for creating Scanned orders.
- `OrderRetrieveDestoryAPIView` for destroying orders.
- `ShiftRetrieveUpdateAPIView` for updating `Shift`s.
- `ShiftAddTimeAPIView`
- `ShiftAddCapacityAPIView`
- `ShiftScannerAPIView`
- `OrderTogglePaidAPIView`
- `OrderToggleReadyAPIView`

- Also added a `JoinShiftAPIView` for joining the `Shift` as a baker via a PATCH to the new endpoint `shifts/<shift>/join`.
- The `ShiftAdminView` now redirects a user to the `JoinShiftView` if he/she has permission to join the `Shift` but is not yet on the bakers list.

Closes #98

### DIFF
--- a/website/orders/api/v1/views.py
+++ b/website/orders/api/v1/views.py
@@ -14,13 +14,14 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from orders import services
-from tosti.api.permissions import HasPermissionOnObject
+from orders.api.v1.permissions import IsOnBakersList
 from orders.api.v1.serializers import OrderSerializer, ShiftSerializer, ProductSerializer
 from orders.exceptions import OrderException
 from orders.models import Order, Shift, Product
 from orders.services import Cart, place_orders, increase_shift_time, increase_shift_capacity
 from tosti import settings
 from tosti.api.openapi import CustomAutoSchema
+from tosti.api.permissions import HasPermissionOnObject
 
 
 class CartOrderAPIView(APIView):
@@ -109,7 +110,7 @@ class OrderListCreateAPIView(ListCreateAPIView):
         """
         Perform the creation of a new Order.
 
-        Permission required: orders.can_manage_shift_in_venue
+        Permission required: orders.can_manage_shift_in_venue and user in shift assignees
 
         This method actually checks if the user has manage permissions in the shift and else ignores the order_type,
         paid and ready fields.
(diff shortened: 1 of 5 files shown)


JobDoesburg (commented):
If we implement it like this, maybe the orders.can_manage_shift_in_venue permission is redundant then?

KiOui:
> If we implement it like this, maybe the orders.can_manage_shift_in_venue permission is redundant then?

No it isn't, this permission is still needed for creating shifts and joining active ones.

JobDoesburg:
Also, does this really close #98? As you 

> > If we implement it like this, maybe the orders.can_manage_shift_in_venue permission is redundant then?
> 
> No it isn't, this permission is still needed for creating shifts and joining active ones.

Yes I see that, but we only need to check that one for creating and joining shifts, for all other views is would be sufficient to just check `IsOnBakingList`

KiOui:
> Also, does this really close #98?

Well, the view actually now returns a redirect but the goal of the Issue is solved.

> Yes I see that, but we only need to check that one for creating and joining shifts, for all other views is would be sufficient to just check `IsOnBakingList`

Yes it would be, but maybe if a user joins the tosticie and then leaves he might still have permission to edit certain shifts as he is on the bakers list. So I guess it would be better to also still check the permission.

website/orders/api/v1/views.py
```diff
@@ -446,3 +480,35 @@ def patch(self, request, **kwargs):
...
+
+    def patch(self, request, **kwargs):
+        """
+        Join Shift as baker view.
+
+        Permission required: orders.can_manage_shift_in_venue
+
+        This adds the requesting User to the Shift assignees.
+        """
+        shift = kwargs.get("shift")
```
JobDoesburg, line 506: you could just add `shift` in the method signature here, so `def patch(self, request, shift, **kwargs):`

KiOui: Yes that is correct, but I did this in the entire codebase so let's fix this in one go sometime.

KiOui: Also: bit stupid of me but I didn't know otherwise.

JobDoesburg (approved):

PR #148, opened 2021-02-20
Buildx for creating ARM images

### LINKED ISSUE #146: ARM docker image in CI
We should add a build for an ARM docker image to the CI so that future ARM based machines (M1, Raspberry Pi's) can also run the docker image natively.

KiOui:
Added `buildx` to the CI for creating ARM docker images.

Closes #146

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -7,7 +7,8 @@ on:
       - master
 
 env:
-  DOCKER_IMAGE: "docker.pkg.github.com/kioui/tosti/tosti-production:latest"
+  DOCKER_IMAGE: docker.pkg.github.com/kioui/tosti/tosti-production:latest
+  DOCKER_TAG_PRODUCTION: docker.pkg.github.com/kioui/tosti/tosti-production:${{ github.sha }}
 
 jobs:
   build-docker:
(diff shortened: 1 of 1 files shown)


KiOui:
Also changed `tosti-production` back to `tosti` for the image name as that is more nice when pulling the image.

.github/workflows/docker.yaml
```diff
@@ -7,7 +7,8 @@ on:
       - master
 
 env:
-  DOCKER_IMAGE: "docker.pkg.github.com/kioui/tosti/tosti-production:latest"
+  DOCKER_IMAGE: docker.pkg.github.com/kioui/tosti/tosti:latest
+  DOCKER_TAG_PRODUCTION: docker.pkg.github.com/kioui/tosti/tosti:${{ github.sha }}
```
JobDoesburg, line 11: If you're doing this, then probably you would just want to build a docker on every PR (with the hash as label), and build :latest on push to master.

KiOui: Yeah I thought this would be nice for downloading "older" images. If you don't agree we can remove it.

JobDoesburg: I don't have a problem with building dockers for each PR! But this action only runs on pushes to master, so I don't see what this does

JobDoesburg: Maybe we should tag `:stable` on release, `:latest` on push to master and `:<commithash>` on each PR?

KiOui: I made up my mind, I'll remove the commit hash push but let's use the stable and latest tags. I think that makes it pretty nice.

JobDoesburg:
So for stable and latest we still need to setup CI

KiOui:
> So for stable and latest we still need to setup CI

Yes

KiOui:
Not for latest though.

JobDoesburg (approved):

PR #151, opened 2021-02-22
User data page

### LINKED ISSUE #96: My order history page
Maybe it is nice to have a page with all orders for a user (also to provide easy access to the data we have about a user)

### LINKED ISSUE #97: My song requests history
Maybe it is nice to see all the song requests of a user

KiOui:
Added two pages under the `/users/account` page that display ordered items and requested tracks to the user.

Also added a `Filter` class within the `tosti` application. This filter class will make it more easier to separate applications as it provides a way to hook onto registered filters that (might or might not) exist. The `Filter` class is inspired by `Wordpress`'s [filter tools](https://developer.wordpress.org/plugins/hooks/filters/).

In the future we can maybe add buttons to clear the users history on-demand.

Closes #96 
Closes #97


website/orders/templates/orders/account_history.html
```diff
@@ -0,0 +1,69 @@
...
+            <li class="nav-item" style="padding-bottom: 0;">
+                <a class="nav-link" id="account-tab" href="{% url 'users:account' %}"
+                   >Personal details</a>
+            </li>
+            <li class="nav-item" style="padding-bottom: 0;">
+                <a class="nav-link active" id="order-history-tab" href="{% url 'orders:history' %}"
+                   >Order history</a>
+            </li>
+            <li class="nav-item" style="padding-bottom: 0;">
+                <a class="nav-link" id="request-history-tab" href="{% url 'thaliedje:history' %}"
```
JobDoesburg, line 22: This violates the app independence

website/users/templates/users/account.html
```diff
@@ -6,19 +6,37 @@
...
-                {% csrf_token %}
-                {{ form }}
-                <input type="submit" value="Save" class="btn btn-primary"/>
-            </form>
+    <div class="container-sm">
+        <h1>Account</h1>
+        <p>The details of your account are displayed below</p>
+        <ul class="nav nav-tabs" role="tablist">
+            <li class="nav-item" style="padding-bottom: 0;">
+                <a class="nav-link active" id="account-tab" href="{% url 'users:account' %}"
```
JobDoesburg, line 14: idem

JobDoesburg (changes requested):

website/orders/templates/orders/account_history.html
KiOui: Yes but how to solve this? This will always happen as we need to create links between the apps.

KiOui: We can solve this with template tags but then still we need to import them in different templates.

website/tosti/filter.py
```diff
@@ -0,0 +1,48 @@
...
+            args_copy = copy.deepcopy(*args)
+            for filter_callback in self.filters[filter_name]:
+                args_copy = filter_callback(args_copy)
+            return args_copy
+        else:
+            return args
+
+
+# The default filter to use
+function_filter = Filter()
```
JobDoesburg, line 48: Let's not call this function_filter, but let's give it a more descriptive name, like `user_data_filter`. Also, maybe it is even better that this filter resides in the `users` app (as it is the users app that creates the users and has the responsibility of providing users with their data)

JobDoesburg: Maybe the `function_filter` can be declared in `users/views.py`, so the other apps will register their data in `user/views.py`

JobDoesburg (changes requested):

website/tosti/filter.py
```diff
@@ -0,0 +1,48 @@
...
+    Generic Filter class.
+
+    This class can create filters and execute them on arguments.
+    """
+
+    def __init__(self):
+        """Initialize."""
+        self.filters = {}
+
+    def add_filter(self, filter_name: str, callback: Callable, place: int = 0):
```
JobDoesburg, line 16: Does this really need a `filter_name`? Can't we just create different filters?

JobDoesburg (approved):

PR #166, opened 2021-08-04
Chore/add django doctor

KiOui:
Added Django-doctor, resolved all issues that Django-doctor pointed out and updated dependencies.


.github/workflows/linting.yaml
```diff
@@ -48,3 +48,6 @@ jobs:
 
       - name: "Run PyDocStyle"
         run: poetry run pydocstyle --explain --add-ignore=D100,D104 --match-dir='(?!migrations).*'
+        
+      - name: "Run Django Doctor"
+        run: poetry run django_doctor check
```
JobDoesburg, line 53: Did you check if this actually fails if errors exist?

JobDoesburg (approved):

Issue #169, opened 2021-09-18
Add reservations application


KiOui:
We should add a reservations application such that people with a BBB (BasisBorrelBrevet) are able to reserve a canteen and drinks/food.

PR #170, opened 2021-10-09
Venue reservations

KiOui:
Added the following:

- A `Reservations` model for the `venues` app
- API endpoints for `venues` app
- Two template tags that can be used for displaying reservations inside a calendar.


KiOui:
Related to #169

PR #172, opened 2021-10-13
Borrel reservations

KiOui:
This pull request adds a `borrel` application to TOSTI which replaces the functions of the [VoorraadCie](https://voorraadcie.nl/) website. The functions of the application include:

- Keeping track of people with a BasicBorrelBrevet
- Enabling people with a BasicBorrelBrevet to create ReservationRequests
- Enabling VoorraadCie members to respond to these ReservationRequests
- Enabling people with a BasicBorrelBrevet to track their ReservationRequests and see whether they are accepted
- Enabling people with a BasicBorrelBrevet to digitally hand in a BorrelForm
- Enabling VoorraadCie members to add and remove Products from the BorrelForm and their inventory

Closes #169

PR #174, opened 2021-10-15
Added a way to request reservations

KiOui:
Added the following:

- An `accepted` field to `Reservation`s which indicates whether the `Reservation` is active/accepted
- Added a form which can be used to request reservations

A way to see `Reservation` requests will be added once we have created the `borrel` application (see #172)

### DIFF
--- a/website/venues/views.py
+++ b/website/venues/views.py
@@ -31,7 +31,7 @@ class RequestReservationView(LoginRequiredMixin, FormView):
 
     def form_valid(self, form):
         """Save the form and add User data."""
-        instance = form.save()
+        instance = form.save(commit=False)
         instance.user = self.request.user
         instance.save()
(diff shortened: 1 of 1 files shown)


KiOui:
> You also want to create a LogEntry for such actions, like https://github.com/svthalia/concrexit/blob/b600592fb6a79383578a2b39186c159ec1822f6d/website/registrations/services.py#L175-L181 does

Where for exactly? Adding a `Reservation` or changing the status of a `Reservation`? The latter is already done by Django right?

JobDoesburg:
> Where for exactly? Adding a `Reservation` or changing the status of a `Reservation`? The latter is already done by Django right?

For changing the status. I think Django will only say "the reservation has changed" but will not display it as "the reservation is approved". Not sure though

JobDoesburg:
Apart from these 3 small comments, it looks good to me

PR #179, opened 2021-10-20
Update to bootstrap 5, fix admin errors

KiOui:
Changed the following:

- Updated Bootstrap 4 to Bootstrap 5
- Removed jQuery
- Changed track search API endpoint, it was echoing an identifier (this was because the old javascript code was a bit stupid, oops)
- Removed Toastr and added TaTa.js (because Toastr uses jQuery)
- Recreated the way we do refreshes on the page, API calls are now not done per function hook but per API endpoint. Functions can register themselves on API data they need. API calls are then done for all functions registered on an endpoint. In the old case, if we had two functions that required the same API data, we did two API requests. Now we only do one and supply the same data to all functions registered on endpoints.

I moved most of the general Javascript functions to the Vue objects as fetch operations are not really easily generalizable.

Test this pull request by testing if all asynchronous (javascript) operations work. Also check the styling of the pages.


JobDoesburg:
One small remark already: we might want to use bootstrap5 offcanvas elements for the styling

JobDoesburg:
afbeelding
The opacity here should not be on the message

JobDoesburg:
afbeelding
there should be padding on the left side of the button here

JobDoesburg:
afbeelding
there should be padding on the left side of the icon here

JobDoesburg:
afbeelding
the close button is not correctly styled

JobDoesburg:
afbeelding
where are the '+5' and 'toggle accept orders' buttons?

JobDoesburg:
afbeelding
this says "join shift" even though I was already joined?

JobDoesburg:
afbeelding
Javascript error

JobDoesburg:
afbeelding
this button does not work

JobDoesburg:
afbeelding
Visiting the calendar is broken

JobDoesburg:
afbeelding
"capacity reached" does not show the current amout of orders (it should display "(1/50)" instead of "(/50)")

JobDoesburg:
afbeelding
This turns out to be possible

JobDoesburg:
After placing an order, the user should be redirected to the shift overview instead of staying at the 'place an order' page

JobDoesburg:
afbeelding
I would expect the order of columns to be reverse

JobDoesburg:
afbeelding
I can't get this to work

JobDoesburg:
afbeelding
I always keep getting a 403 error on every spotify request

JobDoesburg:
afbeelding
not all fields in the inline should be readonly

JobDoesburg:
afbeelding

JobDoesburg:
afbeelding
These should have datetimepickers

JobDoesburg:
afbeelding
hovering this list breaks readability

JobDoesburg:
afbeelding
It should be visible if the shifts are finalized (list_display and list_filter)

KiOui:
Okay so turned out merging with `master` broke everything, #fun.

KiOui:
Why do I have to add migrations to my pull request from the `sp` app?

KiOui:
> afbeelding
> 
> I would expect the order of columns to be reverse

I did not fix this, please create an issue

KiOui:
> afbeelding
> 
> not all fields in the inline should be readonly

This is out of scope for this pull request

KiOui:
> afbeelding

This is also out of scope

KiOui:
> afbeelding
> 
> It should be visible if the shifts are finalized (list_display and list_filter). Also, if the shift is finalized, all fields should be read_only

And this as well

KiOui:
I commented the stuff that I did not change, these things were already in the repository before migrating to bootstrap 5. You can open issues for them if you want to.

JobDoesburg:
I'm still getting only 403's on thaliedje api requests

JobDoesburg:
An overview of other things that should be fixed:

- Change column order of user profile data sections
- Not all shift-admin fields should be read-only
- Fix capitalization in the admin
- list_display and list_filter for finalized shifts
- Product admin broken
- Order admin broken

JobDoesburg (approved):

PR #181, opened 2021-11-29
Prepare for deploy

### LINKED ISSUE #135: Replace OpenID with SAML auth
As [per March 1st 2021, the Science OpenID server will be terminated](https://wiki.cncz.science.ru.nl/Nieuws#End_of_Science_OpenID_service), we need to find a replacement. CNCZ announced that they will be using SAML as alternative, which should be fairly easy for us to implement (https://pypi.org/project/django-saml2-auth/). We only need to wait for CNCZ to respond to our request whether we can use it

JobDoesburg:
Closes #135


requirements.txt
```diff
@@ -0,0 +1,77 @@
...
+charset-normalizer==2.0.8
+click==8.0.3
+commonmark==0.9.1
+coverage==5.5
+cryptography==36.0.0
+defusedxml==0.7.1
+Deprecated==1.2.13
+diff-match-patch==20200713
+Django==3.2.9
+django-admin-autocomplete-filter==0.5
```
KiOui: This file should not be in the repository.

KiOui (approved):

PR #190, opened 2021-12-12
Feature/spotify caching

### LINKED ISSUE #188: Add delay to play and pause buttons after being pressed
Quickly pressing the play and pause buttons for a player does result in a 403 error as the first click causes the spotify player to pause after which the second click can't pause it again.

KiOui:
Changed the following:

- Added caching to Spotify currently_playing, play and pause functionalities
- Added a loader for the player controls when the player is loading data

To test, try out the player and check the caching functionality.

Closes #188

### DIFF
--- a/website/tosti/templates/tosti/base.html
+++ b/website/tosti/templates/tosti/base.html
@@ -6,6 +6,9 @@
     <title>
         {% block title %}TOSTI{% endblock %}
     </title>
+    <script>
+        CSRF_TOKEN = "{{ csrf_token }}";
+    </script>
     {% include 'tosti/analytics.html' %}
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link href="https://fonts.googleapis.com/css2?family=Neucha&display=swap" rel="stylesheet">
@@ -117,9 +120,6 @@
     {% block footer %}
     {% endblock %}
 </div>
-<script>
-    CSRF_TOKEN = "{{ csrf_token }}";
-</script>
 <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/js/bootstrap.bundle.min.js"
         integrity="sha384-MrcW6ZMFYlzcLA8Nl+NtUVF0sA7MsXsP1UyJoMp4YLEuNSfAP+JcXn/tWtIaxVXM" crossorigin="anonymous"></script>
 {% block js %}{% endblock %}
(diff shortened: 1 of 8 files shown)


website/tosti/templates/tosti/base.html
```diff
@@ -6,6 +6,9 @@
     <title>
         {% block title %}TOSTI{% endblock %}
     </title>
+    <script>
+        CSRF_TOKEN = "{{ csrf_token }}";
```
JobDoesburg, line 10: this is not required, your js can just access `Cookies.get('csrftoken')` I think (at least that is what concrexit does)

JobDoesburg (approved):

JobDoesburg:
I'm just thinking that this whole implementation is pretty complex (maybe too complex?). I would personally just have added a "currently playing" and "last time refreshed" attribute to a player. And then in the property based on that, decide if you do the request or not...

JobDoesburg:
Like, wouldn't this be much easier:

```python
class Player(models.Model):
    __currently_playing = None
    __currently_playing_last_refresh = None
    CURRENTLY_PLAYING_CACHE = timezone.timedelta(seconds=2)
    
    (...)


    @property
    def currently_playing(self):
        """
        Get currently playing music information.

        :return: a dictionary with the following content:
            image: [link to image of track],
            name: [name of currently playing track],
            artists: [list of artist names],
            is_playing: [True|False]
        """
        if not self.configured:
            raise RuntimeError("This Spotify account is not configured yet.")

        if self.__currently_playing and self.__currently_playing_last_refresh > timezone.now() - self.CURRENTLY_PLAYING_CACHE:
            return self.__currently_playing

        try:
            self.__currently_playing = self.spotify.currently_playing()
        except JSONDecodeError:
            self.__currently_playing = None
        except OSError:
            self.__currently_playing = None

        if self.__currently_playing is None:
            return False

        image = self.__currently_playing["item"]["album"]["images"][0]["url"]
        name = self.__currently_playing["item"]["name"]
        artists = [x["name"] for x in self.__currently_playing["item"]["artists"]]

        return {
            "image": image,
            "name": name,
            "artists": artists,
            "is_playing": self.__currently_playing["is_playing"],
        }
```

KiOui:
> I'm just thinking that this whole implementation is pretty complex (maybe too complex?). I would personally just have added a "currently playing" and "last time refreshed" attribute to a player. And then in the property based on that, decide if you do the request or not...

I agree that this is a bit more complicated but your suggestion would not work for `play` and `pause` and adding things that we can cache is more complicated with your approach than with mine. If it turns out to be overkill we can revert the changes in the future.

PR #218, opened 2022-03-12
Added Django filter for API endpoints

### LINKED ISSUE #180: Use Django-filter
For a couple of the API views that we have created thus far, it is better to use `django-filters` as described [here](https://www.django-rest-framework.org/api-guide/filtering/). This could be used to merge some API endpoints (such as the search endpoints we have now).

KiOui:
Added API query parameters to the following endpoints:

- Associations list (searching on `name` is now possible)
- Products list (filtering on several parameters is now possible, searching on `name` and `barcode` is now possible), the old `ProductSearchAPIView` endpoint has been deleted in favor of searching on the Products list endpoint. Javascript has also been fixed for the change in endpoint.
- Shifts list (filtering on several parameters is now possible)
- Orders list (filtering on several parameters is now possible)
- Players list (filtering on `venue` is now possible, searching on `display_name` is now possible)
- Venue list (filtering on `can_be_reserved` is now possible, searching on `name` and `slug` is now possible)
- Reservation list (filtering on several parameters is now possible), the old `VenueReservationListAPIView` has been deleted in favor of the normal Reservation list view with a query parameter (e.g. filtering on `venue`). Also note that the `days` parameter for the Reservation list view has been deleted in favor of `start__gte` and `start__lte`. Javascript has also been fixed for the change.

In addition, the following bugs were fixed:

- A bug where `get_queryset` of the ProductsListAPI was called when the schema was generated which raised a Warning.
- A bug where the modal was not being closed after a product was scanned or added in the admin scanner view (due to the upgrade to Bootstrap 5).

Closes #180


website/venues/templates/venues/calendar.html
```diff
@@ -36,7 +36,13 @@
     <script>
         document.addEventListener('DOMContentLoaded', function () {
             let calendarEl = document.getElementById('calendar-all');
-            fetch('{% url 'v1:reservations_list' %}?days=90').then((response) => response.json()).then((data) => {
+            let now = new Date();
+            let yearOffset = (24*60*60*1000) * 365; // 365 days
+            let yearInThePast = now.getTime() - yearOffset;
+            let yearInTheFuture = now.getTime() + yearOffset;
+            let yearInThePastDate = new Date(yearInThePast);
+            let yearInTheFutureDate = new Date(yearInTheFuture);
```
JobDoesburg, line 44: I would refactor these variable names

JobDoesburg (approved):

PR #219, opened 2022-03-13
Recreating how ordes are placed by removing cart orders

### LINKED ISSUE #208: Refactor orders start_date and end_date to start and end

KiOui:
- Refactored the `UserRelatedField` to a `UserSerializer` that is the same for all API endpoints.
- Added an endpoint where a logged-in user can retrieve its details (`/users/me`).
- Moved the `.loader` CSS class to `tosti`.
- Removed the `PlaceOrderView` in favor of buttons on the Order list page.
- Refactored `start_date` and `end_date` to `start` and `end` in `Shift`s.
- Added order buttons to the Order list page.
- Removed the ability to add multiple user orders at once within the `orders` app code.
- Removed the `CartOrderAPIView` together with the `Cart`.
- Removed some duplicate API fields from order and shift endpoint.

Closes #208


KiOui:
I think this is far from actually mergable but I would appreciate it if you gave some feedback now so that I can take that into account before actually merging. You could also alter the interface a bit if you want.

website/orders/templates/orders/order_list.html
```diff
@@ -2,6 +2,27 @@
...
 <div class="container-sm" id="item-orders-{{ shift.id }}">
     <link rel="stylesheet" href="{% static 'orders/css/item-list.css' %}"/>
+
+    <!-- Order buttons -->
+    {% if show_order_buttons %}
+        <div v-if="!shift.finalized && shift.can_order && shift.is_active" class="row row-cols-2 row-cols-md-4">
+            <div class="col mb-3" v-for="product in products">
+                <div class="p-1 d-flex flex-column h-100 rounded" style="background-color: var(--background-shade);">
+                    <p class="m-0 order-name small"><% product.name %></p>
+                    <p class="m-0 item-price small">€<% (Math.round(product.current_price * 100) / 100).toFixed(2) %></p>
```
JobDoesburg, line 12: What is this? What's wrong with decimal fields?

website/orders/templates/orders/shift_header.html
```diff
@@ -5,7 +5,7 @@ <h1>{{ shift.venue.venue.name }}</h1>
         <template v-if="shift.assignees && shift.assignees.length > 0">
             At your service:
             <template v-for="(assignee, index) in shift.assignees">
-                <% assignee %>
+                <% assignee.username %>
```
JobDoesburg, line 8: assignee.username is a username like `s4809327`, so that's not what you want

JobDoesburg (changes requested):

JobDoesburg:
image
these should be centered, the text should be larger and the icon should be added. Also, the whole block should be clickable, not only the add button (though there should be an add-button)

When placing an order, the toast in the upper right corner should not only say "an order was placed" but also what you actually ordered.

website/orders/templates/orders/order_list.html
KiOui: The price is displayed with the least amounts of digits possible, this makes sure that all prices are displayed as they should.

Example: `0,60` is displayed as `0.6` in the API.

website/orders/templates/orders/shift_header.html
KiOui: Yes but not in the API, `__str__` is mapped to `username` in the API serializer for `User`s. I can rename the field if necessary?

KiOui:
> image
> 
> these should be centered, the text should be larger and the icon should be added. Also, the whole block should be clickable, not only the add button (though there should be an add-button). If you want, I can re-design this
> When placing an order, the toast in the upper right corner should not only say "an order was placed" but also what you actually ordered.

This has been changed. I would appreciate it if you changed the interface a bit 😝

KiOui:
Waiting for new data is now done correctly, `OrderSerializer` and `ProductSerializer` are also adjusted to support a POST to the `OrderListCreateAPIView` that lists the required product under the `product` and not under the `product_id`.

website/orders/templates/orders/shift_header.html
JobDoesburg: maybe that is better code quality

JobDoesburg (approved):

JobDoesburg (commented):
I will make some UI changes to the button, and then it can be merged

JobDoesburg:
@KiOui the javascript crashes on orders with no order id, while this is possible

JobDoesburg (approved):

PR #227, opened 2022-06-23
Change user model, remove auto join groups for blacklists and some refactors

JobDoesburg:


website/users/models.py
```diff
@@ -47,65 +46,42 @@ def create_superuser(self, username, **kwargs):
...
     USERNAME_FIELD = "username"
 
-    REQUIRED_FIELDS = []
+    REQUIRED_FIELDS = ["email"]
 
-    class Meta:
-        """Meta class for Users."""
+    username = models.CharField(max_length=8, unique=True)
+    email = models.EmailField(unique=True)
+    full_name = models.CharField(max_length=100)
```
KiOui, line 63: Why do we have `first_name`, `last_name` and `full_name`? Can't we remove either one of these?

KiOui (changes requested):

PR #228, opened 2022-06-28
Fix exception on thaliedje view

KiOui:
Fix an issue where a non authenticated user was not able to acces the Thaliedje now playing view.


JobDoesburg:
Can you also fix this in the orders app?

PR #229, opened 2022-06-28
Fix padding on calendar view

KiOui:
Add padding underneath calendar


JobDoesburg (commented):
Why do this on a separate div and not on the div above?

PR #232, opened 2022-06-29
Added calendar export

KiOui:
Added calendar exports for venue reservations.

### DIFF
--- a/website/venues/views.py
+++ b/website/venues/views.py
@@ -67,9 +67,9 @@ class ListReservationsView(ListView):
 class ReservationFeed(ICalFeed):
     """Output an iCal feed containing all reservations."""
 
-    def product_id(self):
+    def product_id(self, obj):
         """Get product ID."""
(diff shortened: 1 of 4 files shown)


website/venues/views.py
```diff
@@ -60,3 +62,69 @@ class ListReservationsView(ListView):
...
         """Only allow access to user's own reservations."""
         return super().get_queryset().filter(user__pk=self.request.user.pk)
+
+
+class ReservationFeed(ICalFeed):
+    """Output an iCal feed containing all reservations."""
+
+    def product_id(self):
+        """Get product ID."""
+        return f"-//{Site.objects.get_current().domain}//EventCalendar//EN"
```
JobDoesburg, line 72: I think this is supposed to be globally unique: https://www.kanzaki.com/docs/ical/prodid.html

So For the ReservationVenueFeed, I think we should make it a different product id.

JobDoesburg (changes requested):

JobDoesburg (approved):

PR #233, opened 2022-06-30
Feature/filters on explainer page

KiOui:
Added filters for an explainer view and removed separate explainer views.


JobDoesburg:
Very nice

PR #235, opened 2022-06-30
Removed margin from pages and added to header

KiOui:


JobDoesburg:
Look, if I would want to create a page showing an image full-res with only a banner, like this:

```html
{% extends 'tosti/base.html' %}

{% block page %}
    <img src="https://cdn.pixabay.com/photo/2022/06/24/13/57/woman-7281733_1280.jpg" width="100%">
{% endblock %}
```

Then it would look like this:

image


I don't think this should be intended behavior

JobDoesburg:
Maybe we could create a second template `with margin`, that would be acceptable I think

JobDoesburg (approved):

JobDoesburg (approved):

PR #244, opened 2022-07-04
Fixed header and content of some emails

KiOui:
Fixed the header of some emails because SVG is not supported by most email clients. Also fixed the content (URL) of the mails because they were not rendering correctly.


website/borrel/templates/email/borrel_reservation.html
```diff
@@ -1,7 +1,5 @@
 {% extends "email/email.html" %}
 
-{% url "admin:borrel_borrelreservation_change" object_id=borrel_reservation.id as change_url %}
-
```
JobDoesburg, line 4: Why remove this?

JobDoesburg (approved):

PR #261, opened 2022-07-05
Prioritized orders

### LINKED ISSUE #257: Prioritise tosticie members in the list of placed orders
If a non-shift running tosticie member orders a tosti it should move up in the order list.

JobDoesburg:
Closes #257


website/orders/services.py
```diff
@@ -26,6 +27,13 @@ def user_is_blacklisted(user):
     return OrderBlacklistedUser.objects.filter(user=user).exists()
 
 
+def user_gets_prioritized_orders(user, shift):
+    """User's order get put first in the queue."""
+    return user in get_users_with_perms(
+        shift.venue, only_with_perms_in=["gets_prioritized_orders_in_venue"], with_superusers=False
+    )
```
KiOui, line 34: This will return a queryset so you could optimise this with a .filter(pk=user.pk).exists()

website/orders/api/v1/views.py
```diff
@@ -52,7 +52,7 @@ class OrderListCreateAPIView(ListCreateAPIView):
 
     def get_queryset(self):
         """Get the queryset."""
-        return self.queryset.filter(shift=self.kwargs.get("shift"))
+        return self.queryset.filter(shift=self.kwargs.get("shift")).order_by("-prioritize", "created")
```
KiOui, line 55: Don't we want to set this order_by in the `Meta` of `Order`?

KiOui (changes requested):

website/orders/api/v1/views.py
JobDoesburg: I wouldn't, because it will display the ordering differently in the admin.

KiOui: Good point

KiOui (approved):

Issue #287, opened 2022-07-07
Make distinguishing between people easier


wkuijltjes:
If several people with the same name have ordered in the same shift, it is not easy to distinguish between these people, especially if they have ordered the same thing. If everyone's association is shown in the list, this could solve the problem a bit, but if two people of the same association have the same name, perhaps add (the first X letters of) the last name so the tosticie members know whose order is first.

KiOui:
@JobDoesburg I tried adding a [tooltip](https://getbootstrap.com/docs/5.0/components/tooltips/) but this does not work well with VueJS. Then I found a [CSS-only tooltip](https://www.w3schools.com/css/css_tooltip.asp) but this could go wrong with long names (the bootstrap one automatically puts itself where ever it can and the CSS one does not do this). Do you have any other idea on a way of how to solve this or shall I just implement the CSS version?

JobDoesburg:
> @JobDoesburg I tried adding a [tooltip](https://getbootstrap.com/docs/5.0/components/tooltips/) but this does not work well with VueJS. Then I found a [CSS-only tooltip](https://www.w3schools.com/css/css_tooltip.asp) but this could go wrong with long names (the bootstrap one automatically puts itself where ever it can and the CSS one does not do this). Do you have any other idea on a way of how to solve this or shall I just implement the CSS version?

Let's not try css-only. Did you try a [popover](https://getbootstrap.com/docs/5.0/components/popovers/)?

DeD1rk:
It might be an idea to just always display the `display_name` (with ellipsis overflow)...? Or show display name instead of first name when the first name is not unique within the shift of course.

I think that would be more intuitive than hovering/tapping both names to see which is which..

PR #288, opened 2022-07-07
Redirect on first login

### LINKED ISSUE #276: Automatically redirect users to their account profile if no association is set

JobDoesburg:
Closes #276


KiOui:
I think this might also solve #287

KiOui (approved):

PR #289, opened 2022-07-07
Feature/Export borrel application

KiOui:
Added export functionality for the `borrel` application.


website/borrel/admin.py
```diff
@@ -74,10 +148,91 @@ class ReservationItemInline(admin.TabularInline):
     )
 
 
+class BorrelReservationResource(resources.ModelResource):
```
JobDoesburg, line 151: maybe `submitted_at`, `created_at` and `updated_at` are also nice columns to have

JobDoesburg (changes requested):

PR #302, opened 2022-07-11
Allow nu.thalia oauth redirect scheme

DeD1rk:
This way ThaliApp can get credentials using a custom redirect scheme, instead of using https redirects. It's not optimal security-wise, but the risk of some other app intercepting a redirect is small (at least small enough that we allow it on concrexit), and setting up (and testing with) https redirects reliably is a pain.


JobDoesburg:
Why not just redirect to https://thalia.nu and register an endpoint there to open in the app? I don't like our TOSTI codebase depending on thalia in some way

PR #308, opened 2022-07-13
Feature/credits

### LINKED ISSUE #283: Add contributors to site footer
Do a request to https://api.github.com/repos/KiOui/TOSTI/contributors, filter out the accounts with `"type": "Bot", and display them in a muted font somewhere. 

Note that the repo url should be added in settings.py (and also used in https://tosti.science.ru.nl/documentation/)

KiOui:
Adds a config setting that allows an administrator to add a text message to the footer of pages not overwriting it.

Closes #283

### DIFF
--- a/website/tosti/settings/base.py
+++ b/website/tosti/settings/base.py
@@ -168,7 +168,7 @@ CONSTANCE_BACKEND = 'constance.backends.database.DatabaseBackend'
 
 CONSTANCE_CONFIG = {
     'FOOTER_CREDITS_TEXT': ('TOSTI - Tartarus Order System for Take-away Items',
-                            'Text to display in the footer credits, specify [year] for the current year', str),
+                            'Text to display in the footer credits', str),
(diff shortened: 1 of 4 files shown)


website/tosti/settings/base.py
```diff
@@ -166,6 +167,8 @@
 CONSTANCE_BACKEND = 'constance.backends.database.DatabaseBackend'
 
 CONSTANCE_CONFIG = {
+    'FOOTER_CREDITS_TEXT': ('TOSTI - Tartarus Order System for Take-away Items',
+                            'Text to display in the footer credits, specify [year] for the current year', str),
```
JobDoesburg, line 171: How is [year] inserted? 
And does this also work with html?

KiOui: It does not work with HTML, [year] was something I thought of would be nice but let's not do it, it's not needed.

PR #314, opened 2022-07-15
Move orders that are done out of the way

### LINKED ISSUE #258: Move finished tosti orders out of sight
If ordered tosti's are done, it would be nice if these could move down in the list, perhaps enter a second list below the 'processing' list such that the tosti's that still have to be made remain at the top of the screen to prevent unncessary scrolling for the tosticie members running the shift.

### LINKED ISSUE #286: Change the color of additional orders (e.g. sodas/snacks)
If a soda or snack has been ordered, the button to remove it in the order list (admin) is red, whereas the other buttons are all red if something still has to be done (e.g. pay or prepare the tosti), perhaps change this color to make it more clear that its order has been finished.

KiOui:
Created a way in which finished orders are moved from the shift admin screen to a different tab on the shift admin screen.

Closes #258 
Closes #286


KiOui:
@JobDoesburg I would appreciate some feedback on the following:

- Shall we maybe add some kind of animations for orders that will be moved to the other tab (when they are done)?
- It might be a bit unclear that a scanned order is added now because it doesn't show up in the first order tab, what are your thoughts on this?
- I moved the status overview popup to a tab, what do you think about this? Maybe make a different overview (in table format such as the other two tabs)

KiOui:
I solved an issue (stated in the previous message):

- There is now a separate tab with scanned orders, when an admin adds a scanned order, it will be shows with a message and javascript will trigger the scanned orders tab such that it is automatically shown to the user.

JobDoesburg:
A small animation would be nice (just a CSS fade in / out)

JobDoesburg (approved):

KiOui:
I will add an animation via the `TransitionGroup`. This now only works when moving/deleting orders, I will try to get it fixed when adding orders as well.

JobDoesburg:
Looks very good! Maybe implementing ctrl+z woud be nice too

JobDoesburg:
Maybe this is also the time to overwrite the bootstrap default highlight color from blue to tosti-yellow/brown via the css variables

nvoers (approved):

PR #331, opened 2022-08-19
Added ICS borrel reservation export

KiOui:
Added an ICS export for borrel reservations.

To test, visit `/borrel/borrel-reservations.ics`.

### DIFF
--- a/website/borrel/views.py
+++ b/website/borrel/views.py
@@ -1,6 +1,9 @@
+from datetime import timedelta
+
 from django.contrib import messages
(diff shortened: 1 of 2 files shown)


website/borrel/views.py
```diff
@@ -1,13 +1,17 @@
+from datetime import timedelta
```
JobDoesburg, line 1: Doesn't Django time zone have time delta?

KiOui: Yes but we are using `time delta` from `datetime` in all other places in TOSTI.

PR #352, opened 2022-09-21
Add Checkbox for cleaning the canteens when submitting borrel reservation

### LINKED ISSUE #336: Confirmation of following cleaning scheme when submitting a borrel
When Submitting a borrel it's a useful function (for olympus) that you can show that you've cleaned the canteens according to the cleaning scheme.

KiOui:
Added a checkbox when submitting a borrel reservation to check whether people have looked at the cleaning schedule and cleaned the canteens.

Closes #336

### DIFF
--- a/website/borrel/forms.py
+++ b/website/borrel/forms.py
@@ -1,8 +1,10 @@
+from constance import config
 from django import forms
 from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django.forms import models, DateTimeInput
 from django.utils import timezone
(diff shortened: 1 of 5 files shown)


JobDoesburg (changes requested):
I would suggest implementing this in forms.py, see https://github.com/JobDoesburg/PUC-admin/blob/34201e779e6dc8f9730f231c15b82c7ce133f803/pucadmin/frontoffice/forms.py#L30

JobDoesburg (approved):

PR #357, opened 2022-10-04
Add full name on admin order page

### LINKED ISSUE #287: Make distinguishing between people easier

DeD1rk:
Closes #287.

This simple solution should help solve the problem of multiple Max/Tims.

Try out by just looking at the order list.

### DIFF
--- a/website/orders/templates/orders/order_admin_list.html
+++ b/website/orders/templates/orders/order_admin_list.html
@@ -23,30 +23,7 @@
             <div class="tab-pane show active" id="nav-orders-to-process" role="tabpanel" aria-labelledby="nav-orders-to-process-tab">
                 <transition-group name="orders-to-process-list" tag="ul" class="item-orders">
                     <li v-for="(order, index) in orders_to_process" :key="`order_to_process_${order.id}`" class="pt-2 pb-2 item-list">
-                        <div class="information-row">
-                            <p class="item-counter"><% index + 1 %>.</p>
-                            <p class="item-name"><% order.product.name %> (€<% (Math.round(order.order_price * 100) /
-                                100).toFixed(2) %>)</p>
-                            <i v-if="order.product.icon !== null" :class="`fa-solid fa-${order.product.icon} item-icon`"></i>
-                            <div class="flex-divider"></div>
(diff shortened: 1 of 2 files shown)


DeD1rk:
Okay actually this will look pretty strange in some other templates

JobDoesburg:
I would just add it to the order list, maybe with the full last name if not on mobile layout

wkuijltjes:
Adding only the first character of the last name doesn't mean that now all the names on the list are unique, which is more importantly than displaying the first letter of the last name for everybody. E.g. distinguish between Tim van X, Tim van der Y and Tim Z (where Z[0] = V)? Is it too difficult to just check the other users on the list and add ```n``` characters of the last name until a unique user has been created (assuming non-identical names)?

Also, this doesn't resolve the fact that some people might not prefer their official name on the list anyway as they are only known by their nickname (and/or don't like their official name). Displaying a chosen/set ```display_name``` (perhaps in parentheses) might resolve some of the non-unique names issues.

KiOui:
I agree that @wkuijltjes is right about distinguishing the names but I don't think it will be a really big problem as there might not be a lot of same first and last names.

I don't think that we should do a `display_name` because that will definitely be abused (as an example the Thalia website and Jen's photo).

DeD1rk:
Okay I changed it to show first and last name on the shift admin, regardless of any overrides. I think this is fine privacy-wise: tosticie members can see full names anyway in the admin, and it works to distinguish everyone as tosticie member. Users themselves have an icon next to their own orders anyway, so I don't think uniqueness matters there anyway. 

Another note on privacy: the API returns full names to everyone, while not displaying them offers _some_ privacy, it's not like they are then not public any more.

JobDoesburg (approved):

JobDoesburg:
@DeD1rk can you fix the merge conflict

PR #366, opened 2022-10-16
Added day of the reservation to the reservation to fix #358

### LINKED ISSUE #358: Venue reservations that span more than one day don't state this
When a venue is reserved, the homepage displays this as 'currently reserved until XX:XX'. However if the reservation ends not on the current day, it also does this, which could be confusing.
If a venue is reserved from Thursday 10:00 until Friday 14:00, on Thursday it will state 'currently reserved until 14:00', and not state that is continues the next day.

Gulianrdgd:
Closes #358


JobDoesburg:
I would rather only show the day if the day is not today

KiOui:
I agree with Julian, let's just do it like this.

KiOui (approved):

PR #372, opened 2022-11-07
Change date of borrel reservation sync to start date of event

KiOui:
Change the date of a borrel reservation to the start date of the event when synced to Tantalus.


JobDoesburg:
This is very weird because #374 actually wants us to use the prices from the moment of submitting.

PR #382, opened 2022-11-16
Fix slow shift admin

DeD1rk:
Removes the order inlines from the django shift admin entirely, and instead adds buttons to edit a specific order and to create an order to the vue admin.

### DIFF
--- a/website/orders/admin.py
+++ b/website/orders/admin.py
@@ -195,9 +195,6 @@ class OrderAdminForm(forms.ModelForm):
         if "product" in self.fields:
             self.fields["product"].queryset = Product.objects.filter(available=True)
 
-        if "shift" in self.fields:
-            self.fields["shift"].queryset = Shift.objects.filter(can_order=True)
-
         if self.instance.pk:
             if "product" in self.fields:
                 self.fields["product"].queryset = Product.objects.filter(
--- a/website/orders/admin.py
+++ b/website/orders/admin.py
@@ -1,7 +1,6 @@
 from autocompletefilter.admin import AutocompleteFilterMixin
 from autocompletefilter.filters import AutocompleteListFilter
 from django import forms
-
 from django.contrib import admin, messages
 from django.contrib.admin import widgets
 from django.db import models
@@ -9,12 +8,10 @@ from django.db.models import Q
 from django.forms import CheckboxSelectMultiple
 from django.urls import reverse
 from guardian.admin import GuardedModelAdmin
-from import_export.admin import ImportExportModelAdmin, ExportMixin
-from rangefilter.filters import DateRangeFilter
-
-from orders.models import Product, Order, Shift, OrderVenue, OrderBlacklistedUser
+from import_export.admin import ExportMixin, ImportExportModelAdmin
+from orders.models import Order, OrderBlacklistedUser, OrderVenue, Product, Shift
 from orders.resources import ShiftResource
-
+from rangefilter.filters import DateRangeFilter
 from users.models import User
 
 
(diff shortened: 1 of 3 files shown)


website/orders/templates/orders/shift_admin_footer.html
```diff
@@ -41,6 +41,15 @@
                                     <p class="font-footer footer-button-text">Add orders</p>
                                 </div>
                             </div>
+                            <div class="col expand-mobile">
+                                <a target="_blank" :href="`/admin/orders/order/add/?shift=${shift.id}&prioritize=1`">
```
JobDoesburg, line 45: Please don't use a hardcoded url here, but use an url tag

website/orders/templates/orders/order_admin_order.html
```diff
@@ -7,9 +7,12 @@
...
         <div class="ms-auto"></div>
-        <i v-if="order.user !== null && {{ user.id }} === order.user.id"
-        class="fa-solid fa-user user-icon"></i>
-        <p v-if="order.user !== null" class="text-truncate" style="min-width: 0"><% order.user.first_name %> <% order.user.last_name %></p>
+        <p v-if="order.user !== null" class="text-truncate" style="min-width: 0">
+              <i v-if="order.user !== null && {{ user.id }} === order.user.id"
+               class="fa-solid fa-user fa-xs"></i>
+              <% order.user.first_name %> <% order.user.last_name %>
+       </p>
+       <p><a target="_blank" :href="`/admin/orders/order/${order.id}/change/`"><i class="fa-solid fa-pen-to-square fa-xs text-white"></i></a></p>
```
JobDoesburg, line 15: Please don't use a hardcoded url here, but use an url tag

JobDoesburg (changes requested):

website/orders/admin.py
```diff
@@ -79,13 +76,6 @@ class Media:
         """Necessary to use AutocompleteFilter."""
 
 
-class OrderInline(admin.TabularInline):
-    """Inline form for orders."""
-
-    model = Order
```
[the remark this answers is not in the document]
DeD1rk: Yes, even when optimized a lot, there are (many) queries per order. And rendering all of those fields to html (even with non-dropdown widgets for user) is quite slow too. And yes it's really inconvenient, especially since saving a model with the inlines there overrides all changes from the vue admin that happened in the mean time.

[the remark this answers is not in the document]
KiOui: +1

website/orders/templates/orders/shift_admin_footer.html
```diff
@@ -41,6 +41,15 @@
                                     <p class="font-footer footer-button-text">Add orders</p>
                                 </div>
                             </div>
+                            <div class="col expand-mobile">
+                                <a target="_blank" :href="`{% url 'admin:orders_order_add' %}?shift=${shift.id}&prioritize=1`">
```
JobDoesburg, line 45: What does the prioritize field do? Why do we have it?

JobDoesburg (approved):

website/orders/templates/orders/shift_admin_footer.html
DeD1rk: That's the tosticie priority (in GET to pre-fill the checkbox on the add view)

JobDoesburg: Yeah I know, but why do we set it for that button. They just get the priority if they place the order. They don't need to place orders via the admin

DeD1rk: Because this button is there to make it easier for admins to bypass the max number of orders, which we used to do through the django shift admin

Issue #396, opened 2022-12-08
Thaliedje blacklist does not actually blacklist


JobDoesburg:
Skipping is still possible

PR #398, opened 2022-12-10
Venue reservations access and borrel mode

JobDoesburg:


website/thaliedje/api/v1/views.py
```diff
@@ -146,23 +149,22 @@ class PlayerPlayAPIView(APIView):
...
+            return Response(status=status.HTTP_403_FORBIDDEN, data="You are not allowed to control this player.")
+
         try:
             context_uri = request.data.get("context_uri", None)
-            if has_album_playlist_request_permission(request.user, player) and context_uri is not None:
+            if can_request_playlist(request.user, player) and context_uri is not None:
                 services.player_start(player, context_uri=context_uri)
             else:
                 services.player_start(player)
+            log_player_action(request.user, player, "play", "Started playback.")
```
KiOui, line 167: Maybe also log the playlist if it is requested?

website/thaliedje/services.py
```diff
@@ -20,9 +22,64 @@ def user_is_blacklisted(user):
...
+    Get the active ThaliedjeControlEvent for a Player.
+
+    :param player: the player
+    :return: the active ThaliedjeControlEvent or None
+    """
+    qs = ThaliedjeControlEvent.objects.filter(player=player.id, active=True)
+    if qs.exists():
+        return qs.first()
+    return None
+
```
KiOui, line 36: There should only be one right? Maybe throw an exception otherwise.

website/thaliedje/services.py
```diff
@@ -20,9 +22,64 @@ def user_is_blacklisted(user):
...
+
+def can_request_song(user, player):
+    """Check if a user can request a song."""
+    control_event = active_thaliedje_control_event(player)
+    if control_event is not None:
+        if control_event.respect_blacklist and user_is_blacklisted(user):
+            return False
```
KiOui, line 43: Maybe check this in the can_request_song function on the control event as well

KiOui:
One more thing: let's also fix https://github.com/KiOui/TOSTI/issues/396 in this PR.

JobDoesburg:
> One more thing: let's also fix #396 in this PR.

Already did that

website/thaliedje/services.py
JobDoesburg: What do you mean exactly?

JobDoesburg: Wait I understand. Its not possible because of circular dependencies

PR #402, opened 2022-12-16
Add shuffle and replay buttons

### LINKED ISSUE #370: Add shuffle and replay buttons to Spotify client

KiOui:
Add both shuffle and replay button. The replay button only works for `context` replay (so it's not possible from the front-end to set replay to track only). This also adds volume buttons to the slider.

This PR also adds a `debounce` function which can be used to bundle multiple API requests in one go. I have replaced the custom implementation in the search functionality of Thaliedje and have added the `debounce` functionality to the volume slider.

Closes #370


JobDoesburg:
This needs quite a refactor now

website/thaliedje/api/v1/views.py
```diff
@@ -295,7 +295,7 @@ class PlayerShuffleAPIView(APIView):
     )
     serializer_class = PlayerSerializer
     permission_required = "thaliedje.can_control"
```
JobDoesburg, line 297: This has no effect anymore now

website/thaliedje/api/v1/views.py
```diff
@@ -335,7 +338,7 @@ class PlayerRepeatAPIView(APIView):
...
```
JobDoesburg, line 340: idem

JobDoesburg (changes requested):

JobDoesburg (approved):

PR #410, opened 2022-12-28
Add button to place order at bottom of list

### LINKED ISSUE #284: Allow people to be put "on the bottom of the list"
In the case that people have something else to do during the break, people used to put themselves on the bottom of the tosti list but this is no possibility anymore. Perhaps implement a button for this once you have ordered, or allow tosticie members to do this if people have notified the tosticie about this.

It might be generally convenient for tosticie members to be able to move people up or down the list of orders.

nvoers:
Closes #284


website/orders/api/v1/views.py
```diff
@@ -367,6 +367,35 @@ def patch(self, request, **kwargs):
         )
 
 
+class OrderToListBottomAPIView(APIView):
+    """API View to move an order to the bottom of the list."""
+
+    serializer_class = OrderSerializer
+    schema = CustomAutoSchema(response_schema={"$ref": "#/components/schemas/Order"})
+    permission_classes = [IsAuthenticatedOrTokenHasScope]
+    # required_scopes = ["orders:manage"]
```
JobDoesburg, line 376: Why is this commented

PR #438, opened 2023-02-20
Fix new black formatting

KiOui:
Fix the apparently new way black formats things.


JobDoesburg:
Meh I think we first excluded tests or something, but sure

PR #446, opened 2023-03-10
API improvements and interface change

### LINKED ISSUE #435: Term "submit" in list of borrels
Similar to #417.

People find the term "submit" confusing in the list of borrels (Account > view borrel forms). Could this also be changed to "View borrelform"?

KiOui:
This PR improves the API such that it actually becomes usable. The API endpoints for creating and altering Shifts now work as expected and the non-REST-like API endpoints have been removed in favour of a PATCH to the Shift API endpoints. Tests have been added to ensure that the endpoints work as expected.

Also closes #435


website/tosti/api/serializers.py
```diff
@@ -0,0 +1,45 @@
+from rest_framework import serializers
+from rest_framework.serializers import ModelSerializer, ListSerializer
+from rest_framework.utils import model_meta
+
+
+class WritableModelSerializer(serializers.ModelSerializer):
```
DeD1rk, line 6: While this is a cool solution, I feel like it shouldn't be necessary. 
I suppose the problem you're solving is to allow _setting_ a related model, while rendering it with a serializer.
I believe it should actually be possible to do that without this complicated stuff:
- Add the serializer as read only field (nothing special here)
- Also add a `PrimaryKeyRelatedField` with `read_only=False, write_only=True`

I haven't tried it out but that would be a simple and elegant solution. Regarding OpenAPI docs, having the two fields use the same name might be a problem, but again I haven't tried, and working around that probably wouldn't be harder than supporting the WritableModelSerializer

website/orders/services.py
```diff
@@ -131,7 +135,10 @@ def add_user_order(product: Product, shift: Shift, user: User) -> Order:
         type=Order.TYPE_ORDERED,
         user=user,
         user_association=user.association,
+        paid=paid,
+        ready=ready,
         prioritize=user_gets_prioritized_orders(user, shift),
+        deprioritize=deprioritize,
     )
```
DeD1rk, line 144: Can some of the setters below be removed now?

website/tosti/api/serializers.py
KiOui: I thought about this solution and I also think that what you are saying is easier to support. What we do need to do then is to include the normal field as `normal_field` and the primary key field as something like `normal_field_id`. I decided to use this approach because I thought it was more nice to use the `normal_field` for both (to an API user, I think that is more nice to use). @JobDoesburg what do you think?

DeD1rk: ah right, of course it really is impossible to reuse the same name for the two. Any chance it could be done with a custom field type or overriding the serializer to_internal_value?

website/orders/services.py
KiOui: Why?

website/tosti/api/serializers.py
KiOui: Good idea, I will try this!

website/orders/services.py
DeD1rk: Because they're no longer used (e.g. set_shift_active, set_order_ready), so its clean to not leave them behind. Also since the API can now change the models without going through the services, leaving them could falsely suggest that any logic put in them would be adhered to, whereas it will actually never run.

KiOui: Ahh I was looking at other pieces of code, I removed them!

website/tosti/api/serializers.py
KiOui: Although the solution you are proposing is a lot nicer, I have tried implementing this but it seems to be a lot harder than what we have right now. Because of the information that is available after the creation of the class, it is much easier to implement the functionality on the serializer level than on the field level. I propose to use this first and create an issue to look at refactoring this at a later stage.

JobDoesburg:
I'll leave the review to @DeD1rk

website/tosti/api/serializers.py
KiOui: Okay another update: I created a [separate branch](https://github.com/KiOui/TOSTI/tree/fix/interface-changes-b) with a working solution by only using a field. It was tough but it is possible. I however still think that the solution in this PR is easier to understand and a bit cleaner. Could you also give your opinion @DeD1rk ?

DeD1rk: I think doing it as it is now is alright. I have some more ideas that might work (subclass PrimaryKeyRelatedField to override its to_representation, the base field should then handle writing automatically, hopefully leaving only the schema generation as a tricky thing), but it doesn't matter so much. I might try it in a bit but no need to not merge this.

PR #452, opened 2023-03-17
Automatically accept first reservations

### LINKED ISSUE #448: Auto-accepting venue reservations
We also use venue reservations apart from drinks (e.g. for a meeting room). The procedure for accepting these reservations is really easy: If you are the first and it is free, then your reservation is accepted. This is something that could be easily automated (at least from my perspective; but then again, I don't know the back-end programming).

Is is possible to add this feature per venue (a checkbox in the admin panel) such that we can automate the reservations for, in this case, the meeting room?

KiOui:
Closes #448


JobDoesburg (commented):
I personally would have implemented it in the save method and not with a signal, but this is fine

PR #479, opened 2023-05-25
Added OAuth request page

### LINKED ISSUE #472: Page to request API access via OAuth
It would be nice if users could request API access on their own on a TOSTI page (maybe an account tab).

KiOui:
Closes #472


JobDoesburg:
We should at least document that this only supports the authorization code flow. And maybe give some instructions or point to some other documentation on how to use the secret.

Also, the order of the tabs is a bit weird right now

KiOui:
> We should at least document that this only supports the authorization code flow. And maybe give some instructions or point to some other documentation on how to use the secret.
> 
> Maybe also other types of OAuth should be accepted?
> 
> Also, the order of the tabs is a bit weird right now

I've fixed everything but the other OAuth types. We don't want people to be able to create other types as they are either not the most secure (e.g. Implicit) or designed to be resource-to-resource (e.g. we won't have a user associated with them).

PR #493, opened 2023-06-29
Feature/qr code scanner

### LINKED ISSUE #487: Saldo for handed in cans
For a couple of weeks, cans in the Netherlands have gotten "statiegeld". The statiegeld was added to the price of cans in the canteens but there is no way to hand in these cans yet. We should implement some kind of saldo in TOSTI for keeping track of the cans that people have handed in. With this saldo they can then buy new products in the canteens.

### LINKED ISSUE #495: Balance after will be a negative amount when reaching exactly 0 on a transaction

KiOui:
Closes #495 
Closes #487


website/transactions/api/v1/views.py
```diff
@@ -0,0 +1,67 @@
...
+    serializer_class = AccountSerializer
+    queryset = Account.objects.all()
+
+    def post(self, request, **kwargs):
+        """POST handler."""
+        token = request.data.get("token", None)
+        if token is None:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        valid, signed_material = verify_identification_token(token)
```
JobDoesburg, line 34:
```suggestion
        token_valid, username = verify_identification_token(token)
```

website/transactions/api/v1/views.py
```diff
@@ -0,0 +1,67 @@
...
+
+    serializer_class = TransactionSerializer
+
+    def perform_create(self, serializer):
+        """Add request user to validated data."""
+        serializer.save(processor=self.request.user)
+
+    def create(self, request, *args, **kwargs):
+        """Create a Transaction."""
+        if not request.user.is_authenticated:
```
JobDoesburg, line 61: Maybe permission classes are nicer here. Shouldn't we also check OAuth scope? IsAuthenticatedOrTokenHasScope

JobDoesburg (changes requested):

PR #497, opened 2023-09-06
Rewrite VueJS to use script module type

### LINKED ISSUE #491: Rewrite VueJS javascript such that we use script type module

KiOui:
This PR rewrites all VueJS initializers to use the `createApp` function with the `module` script syntax. VueJS is also updated to the most recent version 3.x.

Two things we need to take into account in the future:
- Custom delimiters `<%` and `%>` are no longer supported, I replaced them in favor of `${` and `}$`.
- Because `<script type="module">` blocks do not declare variables in a global scope, I used `window.[variable]` to declare global variables.

To test, open the browser console and run through all the interactive components of TOSTI (so the Shifts and the Thaliedje player).

Closes #491

### DIFF
--- a/website/orders/templates/orders/order_admin_list.html
+++ b/website/orders/templates/orders/order_admin_list.html
@@ -37,9 +37,9 @@
                 <transition-group name="orders-scanned-list" tag="ul" class="item-orders">
                     <li v-for="(order, index) in orders_scanned" :key="`order_scanned_${order.id}`" class="pt-2 pb-2 item-list">
                         <div class="information-row">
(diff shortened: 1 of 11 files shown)


JobDoesburg (approved):
Maybe this is the right time to move the JS to (static) JS files and include those (by loading them in a js post body block), instead of putting them in html templates

KiOui:
> Maybe this is the right time to move the JS to (static) JS files and include those (by loading them in a js post body block), instead of putting them in html templates

Let's make an issue for that 😄

PR #503, opened 2023-09-16
Fix colorblindness for red/green

### LINKED ISSUE #498: Colorblind mode
Jules is colorblind

KiOui:
Closes #498


KiOui:
@JobDoesburg would you mind checking whether you agree with the new shift interface (admin and non-admin)? If you agree I'll update the how-to as well

PR #504, opened 2023-09-18
Yivi integration for 18+ check

KiOui:
This PR adds Yivi integration for an 18+ check.

An abstract `yivi` application is added that can do attribute verification on arbitrary attributes. The `age` application that is added uses the functionality of the `yivi` application to verify the age of a user.


JobDoesburg (approved):
Looks good, but maybe add a field so we can verify if the age registration was done via yivi or manually

KiOui:
> Looks good, but maybe add a field so we can verify if the age registration was done via yivi or manually

Done!

PR #513, opened 2023-10-05
Add time indication for marietje player

KiOui:
This PR adds time indication for the Marietje player. It also fixes a bug where the actual time in minutes and seconds was not shown anymore.


JobDoesburg (approved):
Looks good, but the timeouts in the vue frontend might be better to put in constants instead of hardcoding them

PR #518, opened 2023-11-08
Add needs music keys in reservations

### LINKED ISSUE #455: Notifying Technicie about need for music keys
It would be nice if with a reservation it could also be added that the music keys are needed with a checkbox. Furthermore, it would be nice if the Technicie gets an email about the borrels that tick this box.

KiOui:
Closes #455


JobDoesburg (approved):
I would like it if there's a help_text in the frontend explaining that 

1. you don't need music keys to control the music via Spotify (but that during the reservation, people that reserved the venue can take over the music players and decide who can control the player)
2. if you do use the keys, they should put everything back the way it was set up before (so don't change the master volume)

PR #520, opened 2023-11-12
Add start of status screen and add new picked up button

### LINKED ISSUE #519: Make McDonald's status screen

KiOui:
Closes #519


JobDoesburg (commented):
Looks good. Maybe we should put the status screen in a separate app though. I think that would help keep the orders app small (its already a bit too large)

KiOui:
Shall I merge this?

JobDoesburg:
@KiOui let's separate the status screen

KiOui:
Updated!

website/status_screen/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from status_screen import views
+
+urlpatterns = [
+    path("<shift:shift>/", views.StatusScreen.as_view(), name="status"),
```
JobDoesburg, line 6: Maybe we should add a path with a venue, that automatically redirects to the currently open shift in that venue.

website/status_screen/templates/status_screen/status_screen.html
```diff
@@ -0,0 +1,185 @@
...
+                },
+                refresh() {
+                    if (this.refreshing) {
+                        return;
+                    }
+
+                    clearTimeout(this.refreshTimer);
+                    this.refreshing = true;
+                    return fetch(
+                        `/api/v1/shifts/{{ shift.id }}/orders/?type=0&picked_up=false&ordering=ready_at`,
```
JobDoesburg, line 158: Is there maybe a way to use a {% url %} template tag here?

Like in https://github.com/JobDoesburg/yivi-portal/blob/51ced900cbda534e763f61403eb27eb5251740e6/schememanager/templates/yivi_request.html#L36

JobDoesburg: And similarly on all other places, actually

JobDoesburg (changes requested):

website/status_screen/templates/status_screen/status_screen.html
JobDoesburg: @KiOui status update?

JobDoesburg (approved):

PR #521, opened 2023-11-15
Add daily opening requirement

JobDoesburg:
Kijk eens, dit is elegante code

### DIFF
--- a/website/fridges/services.py
+++ b/website/fridges/services.py
@@ -29,10 +29,11 @@ def user_can_open_fridge(user, fridge):
     if not opening_hours.exists():
         return False, None
 
-    already_opened_today = AccessLog.objects.filter(fridge=fridge, timestamp__date=timezone.now().date()).exists()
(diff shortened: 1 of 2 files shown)


website/fridges/services.py
```diff
@@ -24,10 +26,14 @@ def user_can_open_fridge(user, fridge):
...
 
     opening_hours = fridge.current_opening_hours
-
     if not opening_hours.exists():
         return False, None
 
+    already_opened_today = AccessLog.objects.filter(fridge=fridge, timestamp__date=timezone.now().date()).exists()
+    # This requires a daily opening by a user with the open_always permission
+    if config.FRIDGE_REQUIRE_DAILY_OPENING and not already_opened_today:
+        return False, None
```
KiOui, line 35: Let's only compute the `already_opened_today` if the feature flag is on (this saves a database call when the feature flag is off).

KiOui (changes requested):

KiOui (approved):

PR #530, opened 2023-12-13
add deprioritize function for bakers

### LINKED ISSUE #528: Deprioritize option for Tosticie
After shouting someone's name two, or maybe three times, it would be nice to be able to push them down the list by deprioritizing. This way the Tosticie don't have to keep shouting their names until they finally show up, but they have to wait a bit for their tosti.
(This could make #527 redundant)

mark-boute:
closes #528

order_admin_list.html
- copied this function here, we could choose to move it to the shift header/footer files or make a function template but that felt like overdoing it

order_admin_order.html
- line 18, 19 and 20 are new, 18 is a bit ugly but was needed for keeping the col spacing uniform. Alternatives were more complex

### DIFF
--- a/website/orders/templates/orders/order_admin_list.html
+++ b/website/orders/templates/orders/order_admin_list.html
@@ -314,14 +314,14 @@
                         }
                     ).then(response => {
                         if (response.status === 200) {
-                            return response;
+                            return response.json();
                         } else {
                             throw response;
                         }
                     }).then(() => {
+                        tata.success('', 'Deprioritized order successfully');
                         if (typeof (update_refresh_list) !== 'undefined') {
                             update_refresh_list();
-                            tata.success('', 'deprioritized successfully`);
                         }
                     }).catch(error => show_error_from_api(error));
                 }
--- a/website/orders/templates/orders/order_admin_list.html
+++ b/website/orders/templates/orders/order_admin_list.html
@@ -321,6 +321,7 @@
                     }).then(() => {
                         if (typeof (update_refresh_list) !== 'undefined') {
                             update_refresh_list();
+                            tata.success('', 'deprioritized successfully`);
                         }
                     }).catch(error => show_error_from_api(error));
                 }
--- a/website/orders/templates/orders/order_admin_list.html
+++ b/website/orders/templates/orders/order_admin_list.html
@@ -296,6 +296,35 @@
                     })
                 });
             },
+            order_to_bottom_of_list(order) {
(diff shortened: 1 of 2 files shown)


KiOui:
Hey Mark! Thanks for the PR, it looks nice! I would like to see some changes however:

- Let's not use the same buttons as the "Paid" and "Ready" buttons, let's make them an icon just like on the normal order page (so just the font awesome icon).
- It's nice to give the user a toast feedback when the action succeeds/fails. So we should add a `tata.success` or `tata.error` message on fail.

website/orders/templates/orders/order_admin_list.html
```diff
@@ -296,6 +296,35 @@ <h3 class="mt-3">Orders total</h3>
...
+                    ).then(response => {
+                        if (response.status === 200) {
+                            return response;
+                        } else {
+                            throw response;
+                        }
+                    }).then(() => {
+                        if (typeof (update_refresh_list) !== 'undefined') {
+                            update_refresh_list();
+                        }
```
KiOui, line 324: So below this, add a `tata.success` with a success message (you can also see how this works on the normal order page).

website/orders/templates/orders/order_admin_order.html
```diff
@@ -11,17 +11,22 @@
...
-           <button v-else type="button" class="checkbox-paid btn col btn-danger"
-                   v-on:click="toggle_paid(order)">Paid <i class="fa-regular fa-circle-xmark"></i></button>
-           <button v-if="order.ready" type="button" class="checkbox-ready btn col btn-success"
-                  v-on:click="toggle_ready(order)">Ready <i class="fa-regular fa-circle-check"></i></button>
-           <button v-else type="button" class="checkbox-ready btn col btn-danger"
-                  v-on:click="toggle_ready(order)">Ready <i class="fa-regular fa-circle-xmark"></i></button>
-       </div>
+        <button v-if="order.deprioritize" type="button" class="btn col" style="color: transparent;" disabled></button>
+        <button v-else type="button" class="btn col btn-warning" v-on:click="order_to_bottom_of_list(order)">
+            <i class="fa-solid fa-arrow-down-short-wide" style="font-size: 1.5em; cursor: pointer;"></i></button>
```
KiOui, line 20: Let's not use the same buttons as for "Paid" and "Ready" but let's add a clickable font awesome icon. @JobDoesburg or do you have any other good suggestion?

KiOui (changes requested):

KiOui:
@mark-boute I've fixed some of the issues that were still remaining and I'm going to approve this PR, thanks for the work!

PR #531, opened 2023-12-13
explitictly unset prioritize on deprioritizing

### LINKED ISSUE #525: Prioritise/deprioritise conflict
When an order has priority (because someone is a tosticie member for example) and you deprioritise your own order, the order stays on top of the order list, since both the prioritise and deprioritise are on. It would be nice if these could be exclusive, meaning the order is either prioritised or deprioritised, and not both.

mark-boute:
closes #525
Make sure that the prioritize flag is unset on de-prioritizing.

### DIFF
--- a/website/orders/api/v1/views.py
+++ b/website/orders/api/v1/views.py
@@ -133,7 +133,8 @@ class OrderRetrieveUpdateDestroyAPIView(LoggedRetrieveUpdateDestroyAPIView):
                 if instance.deprioritize and not deprioritize:
                     self.permission_denied(request, message="Deprioritize can not be unset.", code=400)
                 else:
-                    serializer = self.get_serializer(instance, data={"deprioritize": deprioritize}, partial=True)
+                    data = {"deprioritize": True, "prioritize": False} if deprioritize else {"deprioritize": False}
+                    serializer = self.get_serializer(instance, data=data, partial=True)
                     serializer.is_valid(raise_exception=True)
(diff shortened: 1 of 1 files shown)


mark-boute:
@KiOui any idea why this is not unsetting the flag currently? I am stuck

KiOui:
> @KiOui any idea why this is not unsetting the flag currently? I am stuck

The `OrderSerialize` has the `prioritize` field as read-only. It should be altered in such a way that `deprioritize` overrules `prioritize`. I think we have three options here:

- The best option is to just create a new field "priority" with three options: "deprioritize", "normal" or "prioritize". This option does however require a database migration.
- We could also maybe adjust the `save` method to automatically set `prioritize` to `false` whenever `deprioritize` is `true`.
- Or we could change the `OrderSerializer` to also accept the `prioritize` field and then adjust the `OrderRetrieveUpdateDestroyAPIView` such that it only accepts changes to the `prioritize` field if the user has permission.

JobDoesburg:
For me, this is a frontend thingy. There's nothing wrong with both prioritize and deprioritize flags set in principle. It's just not the behavior people expect if they hit the button. So keep it in the API / serializer

KiOui:
I don't agree with you on that point. It's weird to have an order that both has priority and does not have priority. So I think migrating this to just one field that can either be prioritized, normal or deprioritized is the best option.

KiOui:
This has been fixed in #534

PR #534, opened 2023-12-22
Replace prioritize with one field

### LINKED ISSUE #525: Prioritise/deprioritise conflict

KiOui:
This PR replaces the `prioritize` and `deprioritize` fields with one `priority` field that can either be `prioritize` or `deprioritize` (or `normal`).

Closes #525


KiOui:
@mark-boute This fixes the thing discussed in #531

mark-boute (approved):
Nice, much more modular too :D

PR #578, opened 2024-10-25
Add status screens for order venues status screens

vincevd1:
This adds a status screen that automatically switches between the original status screen for shifts and a music screen to see what is currently playing.

From what I have seen you do not get logged out of the current session as long as the browser stays open which is the case with thaliedje. So we can just log in with an account with no privileges and put this link on. 

I hooked up a monitor in noord to show this screen.

### DIFF
--- a/website/status_screen/templates/status_screen/venue_music_screen.html
+++ b/website/status_screen/templates/status_screen/venue_music_screen.html
@@ -1,0 +1,234 @@
+{% extends 'tosti/base.html' %}
+{% load static players %}
+
+{% block styles %}
+    <link rel="stylesheet" href="{% static 'status_screen/css/status-screen.css' %}"/>
+    <link rel="stylesheet" href="{% static 'thaliedje/css/player.css' %}"/>
+{% endblock %}
+
+{% block header %}
+    <nav class="navbar navbar-expand-lg site-header sticky-top navbar-dark">
+        <div class="container">
+            <div class="d-block d-lg-none">
+                <a class="navbar-brand drop-out-header-mobile" href="/"><img
+                        src="{% static 'tosti/svg/TOSTI-logo.svg' %}" height="80"/></a>
+            </div>
+            <div class="mx-auto position-relative d-lg-block d-none" style="margin-top: -68px">
+                <a class="navbar-brand drop-out-header" href="/"><img
+                        src="{% static 'tosti/svg/TOSTI-logo.svg' %}"
+                        height="120"/>
+                </a>
+            </div>
+        </div>
+    </nav>
+{% endblock %}
+
+{% block page %}
+    <div class="d-flex flex-column justify-content-center align-items-center mt-5" id="music-screen-container">
+        <h1 class="mt-5">Currently playing:</h1>
+        <div class="player" id="player-container" style="width: 65%">
+        <template v-if="player_data.track.name !== null">
+        <div class="w-50 mx-auto mb-2" v-if="player_data.track.image !== null">
+            <a href="{% url "thaliedje:now_playing" player=player %}">
+                <img class="w-100" :src="player_data.track.image" :alt="player_data.track.name"/>
+            </a>
+        </div>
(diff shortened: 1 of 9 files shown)


website/status_screen/templates/status_screen/venue_music_screen.html
```diff
@@ -0,0 +1,234 @@
...
+                        ${ artist }$ -
+                    </template>
+                </template>
+            </h3>
+        </div>
+        </template>
+        <template v-else>
+            <h1 class="text-center m-auto">No currently playing track</h1>
+        </template>
+    </div>
```
JobDoesburg, line 74: Can't we simply include the player card here? We have template_tags to render a player...

JobDoesburg (changes requested):
Code looks fine, will have to see it live to say something about the styling

website/status_screen/templates/status_screen/venue_music_screen.html
vincevd1: Yeah I saw that and I tried, but since my page contains vue code and the template does as well this does not go well. The vue on my page skips over the script that contains the vue code for the template tag. And I get the following error:

image

So I tried to copy as little as I need but I will try to look for a better alternative since this is a lot of double code.

vincevd1: Nvm I might just be a fucking idiot and I should put the template outside of the component I mount the other vue on...

vincevd1:
Styling of the music screen looks like the following:

image

Large text because it has to be able to be seen on the monitor from far away

KiOui:
Hey Vince! Thanks for the MR. I have two suggestions:

- I've created OAuth2 support for one of the status screens, you can find the PR [here](https://github.com/KiOui/TOSTI/pull/579).
- I think it would be better if we would put this all into one screen without the need for refreshing. I would do it the following way:

1. Check if a music player exists, if it does, tell the template to also render a music player.
2. In the template, check if an active shift exists, if it does, display and refresh that.
3. If no shift is active, switch over to displaying the music player.

I think this is cleaner than refreshing the page (I also got into an infinite refresh loop when testing).

website/tosti/static/tosti/js/base.js
```diff
@@ -162,5 +162,37 @@ function formatPrice(priceToFormat) {
     }).format(priceToFormat).trim();
 }
 
+function get_oauth_token_with_client_credentials(client_id, client_secret) {
+    let credentials = client_id + ":" + client_secret;
+    let credentials_utf8 = strToUTF8Arr(credentials);
+    let credentials_base64 = base64EncArr(credentials_utf8);
+    return fetch(
+        "/oauth/token/",
```
JobDoesburg, line 170: Let's not hardcode this

KiOui:
@vincevd1 I made some small changes again. I removed the dedicated active shift endpoint in favor of using the list shift endpoint with query parameters

website/status_screen/templates/status_screen/venue_status_screen.html
```diff
@@ -0,0 +1,275 @@
...
+    <div id="player-container" style="display: none">
+        {% if player is not None %}
+            <div class="player w-75 mt-5 m-auto display-5">
+                <h1 class="mt-5">Currently playing:</h1>
+                {% render_player player %}
+            </div>
+        {% else %}
+            <h1 class="mt-5">No player for this venue</h1>
+        {% endif %}
+    </div>
```
KiOui, line 71: You don't have to do this with `display: none`, you can show this with another template and then adding `v-else` to the `v-if="active-shift"` container above.

KiOui (changes requested):
Nicely done @vincevd1! Just one small thing and we can merge this 😄 You may also overwrite the old status screen if you want to.

website/status_screen/templates/status_screen/venue_status_screen.html
vincevd1: I tried that but it gives me an error because we are already mounting a vue app on the container above, that's why I used this hack

KiOui (approved):

PR #579, opened 2024-10-26
OAuth2 support for status screen

KiOui:
You can enable OAuth2 by setting `TOSTI_CLIENT_ID` and `TOSTI_CLIENT_SECRET` cookie values to values retrieved from the admin. Create a Client Credentials OAuth application and copy over the ID and secret.


vincevd1:
Okay seems good, I will merge this and apply the changes you suggested to get everything in one file and avoid refreshing the page

PR #580, opened 2024-11-15
Allow admin for shift without joining it

### LINKED ISSUE #533: Een shift kunnen joinen zonder dat je hebt gebakken om orders af te kunnen ronden en de shift te finalizen

KiOui:
Closes #533 

This PR alters the `JoinShiftView` to allow a user to go to the admin page without being in the assignees. The user will get redirected to the `JoinShiftView` once every 10 minutes if accessing the shift admin view without being in the assignees.


vincevd1 (commented):
LGTM, however I do think it might be a good idea to update the templates a little bit to make clear it is also possible to only manage. Right now just pressing no when wanting to join might be a little bit unclear.

PR #585, opened 2025-02-19
Add BBB requirement to certain venues (all venues except kees)

vincevd1:
Olympus requested for an option to make canteen reservations impossible for people without a BBB.
This commit implements that.


KiOui:
This change is not possible in this way. We have a small problem here maybe @JobDoesburg knows a nice solution.

We don't want any circular dependencies in TOSTI but with this change we create circular dependencies because `borrel` depends on `venues` and (with the changes here) `venues` depends on `borrel`. We should probably thus move `BasicBorrelBrevet` out of `borrel` into a new app. @JobDoesburg what do you think of that?

Additionally, @vincevd1 I think the check needs to be extracted from the form into the view. The form checks usually do not check permissions of users, we should do that in the forms. You should also adjust the checks in the API endpoints. Don't forget that a borrel reservation can also trigger a venue reservation so there should probably also be a check made there.

JobDoesburg:
RT Lars. Because the check is enforced in the view, not in the model or admin it doesnt actually crash right now, but the dependency is circular which will cause problems at some point

vincevd1:
Ah yeah that is a good point. In that case I think it would also be better to move BasicBorrelBrevet. Maybe it can be the attribute of a user so we move it into the user app?

I don't really understand what you mean with we should enforce the permission check in the view instead of the form. Do you mean when the user submits the form? Or in `forms.py` where I check permissions to know which venues to show. If I have to remove it from `forms.py` how would I hide some venues from people without a BBB?

I did forget to implement the check in the API and will change this soon. Also I don't think we have to worry about the borrel reservation triggering a venue reservation since you need a BBB for a borrel reservation anyway.

KiOui:
I have rebased this MR for you. Specifically, the following changes are still required:

- [ ] A permission check should be added to the reservation create API endpoint. You can do that in `check_permissions` of `ReservationListCreateAPIView`. Note that retrieving reservations should still be possible for all users.
- [ ] The `Add Reservations` view should not be viewable if the user has no basic borrel brevet and all venues require one.
- [ ] The `Add Venue Reservation` button should not be displayed if the user has no basic borrel brevet and all venues require one.

KiOui:
You can fix the `test_forms.py` tests by supplying a `request` keyword argument that contains a user (as mock).

PR #588, opened 2025-03-04
Move BBB to certificates app

KiOui:
## Test instructions

1. Create a user, add a borrel brevet.
2. Create some borrel products and a venue.
3. Try to create a borrel reservation with the user.
4. Try to finish a borrel reservation.


JobDoesburg:
Not too enthusiastic about the name `certificates`. Perhaps it would be fun to integrate it with `age` as some sort of `qualifications` you can store about people?

KiOui:
> Not too enthusiastic about the name `certificates`. Perhaps it would be fun to integrate it with `age` as some sort of `qualifications` you can store about people?

I've renamed the app to `qualifications`. I've not included the `age` app yet, if necessary we can do that in the future.

vincevd1:
I added the implementation of the requirement for the BBB from the other branch to this branch and also implemented into the API.

KiOui:
> I added the implementation of the requirement for the BBB from the other branch to this branch and also implemented into the API.

I think it's nicer to merge this first and then work on your thing. This is a different change 😄

PR #615, opened 2026-04-22
Switch from django-cron to celery

KiOui:
Move from the `django-cron` application to Celery for asynchronous tasks


deploy/docker-compose.yml
```diff
@@ -40,6 +40,12 @@ x-django-web-env: &django-web-env
   DEFAULT_FROM_EMAIL: "TOSTI <${EMAIL_ADDRESS}>"
   SERVER_EMAIL: ${EMAIL_ADDRESS}
 
+x-django-celery-env: &django-celery-env
+  <<: *db-env
```
JobDoesburg: Shouldn’t we have the Django core env too

deploy/docker-compose.yml
```diff
@@ -134,9 +140,33 @@ services:
...
+      - media:/media
+      - cache:/app/cache
+    environment: *django-celery-env
+
+  celery_beat:
+    # This container will run celery beat (cron) tasks.
     image: *tosti-image
-    entrypoint: ["/bin/sh", "/entrypoint_cron.sh"]
+    entrypoint: [ ]
+    command: celery --app=mode_groothandel beat --loglevel=info --uid=nobody --scheduler django_celery_beat.schedulers:DatabaseScheduler
```
JobDoesburg: Groothandel?

deploy/docker-compose.yml
```diff
@@ -147,21 +177,24 @@ services:
...
-      # migration and crash with "relation does not exist" — which lands
-      # in Sentry on every deploy.
       web:
         condition: service_healthy
     volumes:
       - media:/media
       - cache:/app/cache
-    environment: *django-core-env
+    environment: *django-celery-env
+  redis:
```
JobDoesburg, line 187: Newline please

JobDoesburg (changes requested):
Also, where do we schedule when this runs? We only say "shared_task" but never when and how frequently the tasks should run?

deploy/docker-compose.yml
KiOui: Ahh yes you are right!

KiOui:
> Also, where do we schedule when this runs? We only say "shared_task" but never when and how frequently the tasks should run?

We can schedule the tasks in the database when we have deployed.

deploy/docker-compose.yml
KiOui: Ahh I absolutely did not copy this over.

deploy/docker-compose.yml
```diff
@@ -147,21 +155,47 @@ services:
...
       - cache:/app/cache
-    environment: *django-core-env
+    environment: *django-celery-env
+
+  celery_beat:
+    # This container will run celery beat (cron) tasks.
+    image: *tosti-image
+    entrypoint: [ ]
+    command: celery --app=tosti beat --loglevel=info --uid=nobody --scheduler django_celery_beat.schedulers:DatabaseScheduler
+    restart: unless-stopped
```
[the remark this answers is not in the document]
KiOui: We are going to manually create these schedules.

website/silvasoft/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+
+    maximum_synchronizations_to_run = config.MAXIMUM_SYNC_PER_RUN
+    succeeded = 0
+    failed = 0
+
+    borrel_reservations_to_synchronize = BorrelReservation.objects.filter(
+        silvasoft_synchronization__isnull=True,
+        submitted_at__isnull=False,
+    ).order_by("submitted_at")
+
```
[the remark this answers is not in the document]
KiOui: We should actually queue a task here when a shift is made final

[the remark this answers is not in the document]
KiOui: Tasks are now queued when a `Shift` is finalised and when a `BorrelReservation` is submitted. I also added tests to verify whether the `pre_save` and `post_save` methods actually trigger the behaviour we want.

PR #631, opened 2026-05-27
Menstruacie Tampon feature

Luukosaurus:
This Project was made by:
Widad Majdoubi,
Thomas Jansen and
Luuk Schukkink

We made a page on the site that can be used to notify the Menstruacie for restocks and or repairs of the menstrual product dispensers.

**Key features**

Front page where you can submit a notification.
Second page for committee members to see and resolve these notifications.

Rooms and stocks need to be created in the admin before use, and notifications and restocks can be managed there as well.

We made a permission that needs to be assigned to the Mentruacie. With this permission, members can only use the notifications page and can not edit or see anything in the admin. For this, the default Django permissions can be used.

The Front page also has support for QR codes, which can be tested by using ?room=hg01002 for it to automatically fill in room hg.01.002 if it exists.

**For testing**

You need to add at least one room to test the user form and its QR-code function.
You need one unresolved notification and a "Stockdata" to test the notification page and its resolve function.


website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        same_room = Room.objects.filter(
+            building=self.building,
+            floor_number=self.floor_number,
+            room_number=self.room_number,
+        )
+        same_room = same_room.exclude(pk=self.pk)
+        if same_room.exists():
+            raise ValidationError(
+                "A room with this building/floor/room number already exists."
+            )
```
KiOui: You can do this with a `UniqueConstraint`.

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        return None
+
+    def save(self, *args, **kwargs):
+        self.clean()
+        building = self.BUILDING_CODES.get(self.building, "xx")
+        self.room_code = f"{building}{self.floor_number:02d}{self.room_number:03d}"
+        super().save(*args, **kwargs)
```
KiOui: I don't like this. The integers that you use here can be longer than respectively 2 or 3 digits. Can't we just make the `room_code` something that can be computed from it's properties and put a restriction on the amount of digits `floor_number` and `room_number` can have?

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __str__(self):
+        return f"{self.building} {self.floor_number:02d}.{self.room_number:03d}"
```
KiOui: So we could for example use this method for computing the `room_code` on demand without saving it in the database as duplicate information.

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        max_length=500,
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    is_resolved = models.BooleanField(default=False)
+    resolved_at = models.DateTimeField(null=True, blank=True)
+
+    class Meta:
+        ordering = ["-created_at"]
+        verbose_name = "Tampon notification"
+        verbose_name_plural = "Tampon notifications"
```
KiOui, line 32: I think these will be set automatically right?

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        permissions = [
+            ("manage_tamponnotification", "Can manage Tampon notifications"),
+        ]
+
+    def mark_resolved(self):
+        self.is_resolved = True
+        if self.resolved_at is None:
+            self.resolved_at = timezone.now()
+        self.save(update_fields=["is_resolved", "resolved_at"])
```
KiOui: Let's add this behaviour in the `save()` method.

website/tampon/permissions.py
KiOui, line 1: Why do we need these "special" permissions? Can't the menstruacie just look at the notifications in the admin view? Then you would just be able to use the default Django permissions and you won't have to create a frontend view for viewing the notifications.

website/tampon/templates/tampon/notifications.html
KiOui, line 1: So I think it is nicer to create this view in the admin instead of on the frontend. See my comment about the permissions above.

KiOui (changes requested):
I'm currently not able to do a full review, but here is already some feedback. First of all: This is a very nice PR and very nice idea, I also really like the committee name and the application name!

website/tampon/templates/tampon/index.html
```diff
@@ -0,0 +1,106 @@
...
+T.A.M.P.O.N.
+{% endblock title %}
+{% block page %}
+<div class="container mt-5">
+    <div class="mb-5 text-center">
+        <h1>
+            T.A.M.P.O.N.
+        </h1>
+        <p class="lead tagline">
+            Tracker for Availability of Menstrual Products & Out-of-stock Notifications
```
JobDoesburg: What about adding `<i class="fa-solid fa-circle fa-xs me-4 align-middle" aria-hidden="true"></i>`
Or is that a bit too much haha

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        "Huygens": "hg",
+    }
+
+    building = models.CharField(
+        max_length=10, choices=BUILDING_CODES, default="Huygens"
+    )
+    floor_number = models.IntegerField()
+    room_number = models.IntegerField()
+    active = models.BooleanField()
+    room_code = models.CharField(max_length=7, unique=True, editable=False)
```
JobDoesburg: Is it really desirable to store these separately? Is it really desirable to filter on floor number, room number and building separately? 
Since it's just one building, and I cant imagine there are so many rooms, this feels a bit like premature optimization. 

I think its much simpler and more convenient to just have rooms with 2 fields, a `display_name` and a `slug`.

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        self.clean()
+        building = self.BUILDING_CODES.get(self.building, "xx")
+        self.room_code = f"{building}{self.floor_number:02d}{self.room_number:03d}"
+        super().save(*args, **kwargs)
+
+    def __str__(self):
+        return f"{self.building} {self.floor_number:02d}.{self.room_number:03d}"
+
+
+class TamponNotification(models.Model):
```
JobDoesburg, line 17: I'm afraid this is a bit feature creep. Do we really need to keep track of notifications? And are we really gonna update the `is_resolved` field? 

It is much simpler if we would just send an email when somebody makes a notification, and not store anything in TOSTI. Maybe that's even more convenient for the committee members than checking this website every once in a while. Wouldnt that suffice?

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+    """Notification submitted through the tampon form."""
+
+    room = models.ForeignKey(Room, on_delete=models.CASCADE)
+    notification_text = models.TextField(
```
JobDoesburg: Do we really need a notification text?

website/tampon/models.py
```diff
@@ -0,0 +1,115 @@
...
+        blank=True,
+        max_length=500,
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    is_resolved = models.BooleanField(default=False)
```
JobDoesburg, line 26: Are people actually gonna update this field?

JobDoesburg (changes requested):
Some mechanical and cosmetical remarks from me. Most important point, however, is about the scope. Do we really want to keep track of notifications, restockings, etc etc in the TOSTI database? If yes, that's fine by me, but from what I heard about the project earlier, the idea was just to send out an email. So then this whole schema is a bit feature creep.

KiOui:
Also: The idea is to hang up posters in the bathrooms with QR codes that people can scan right? Would it be nice to make a QR code in the admin backend of a room that can be printed out?

website/tampon/models.py
KiOui: Or, as @JobDoesburg said below, make the room code a slug with a displayable field.

JobDoesburg:
> Also: The idea is to hang up posters in the bathrooms with QR codes that people can scan right? Would it be nice to make a QR code in the admin backend of a room that can be printed out?

That really sounds like feature creep 🙂

website/tampon/permissions.py
Luukosaurus: I'm currently working on fixing the issues in our code but this is more of a design choice. Because Thomas preferred it that committee members don't have unneeded staff status or permissions and this makes it really easy to use for non technical committee members. If you really think this is a bad design choice we are open to discuss this.

website/tampon/models.py
Luukosaurus: I agree that for example room_code can be removed but active is needed and after that 
Building Floor_number Room_number are used to create this slug and display_name. So I don't think this 3 to 2 ratio is that big of a waste compared to the benefit of the ease of use gained by it.

But I will look in to this and we will discuss this and at least try to remove room_code.

website/tampon/models.py
Luukosaurus: I think sending an email is a good idea but as far as I know the committee wanted to store the data for future data analysis on it. Especially the restocks are useful to see its effectiveness.
I'm not in charge of design so this email part is not my call but i will communicate this to the committee.

I personalty think it might be nice to just delete the notification after a while but again this is partly a design decision so I will ask the committee.

website/tampon/models.py
Luukosaurus: This is used so people can indicate "Hey the dispenser is broken" or if they want be extra helpful and state which product is empty. This seems easier for the committee and for the user than making every product a field or making vandalism/damage a extra field.

website/tampon/models.py
Luukosaurus: I'm not sure what you mean. Created at is never changed again in future but its nice to know when its created.
And is_resolved is changed only once and then under normal usage never again.

website/tampon/models.py
JobDoesburg: Yes I agree that `active` should be there, was mainly about room+floor+code and the whole logic of deriving the display name from them. Just having `slug` and `display_name` (and active) is much more versatile. And for example helps you to later specify things like "HG00.123a (left)" if they are ever needed (which will probably the case). Or other cases where boxes may be at locations that turn out to not follow the room number conventions. (What if one will be placed in a hallway)

website/tampon/models.py
JobDoesburg: Yep, please discuss this. The promise of "we'll run some data analysis" often turns out as something that "might be nice for the future but is never going to happen" in my experience. And then it would be a shame to put in the effort in a system that is not going to be used. Or worse: a system that doesnt satisfy the users (because it keeps showing notifications until they are dismissed/restocked for example, making people frustrated and deciding to step away from it anyway.

As per Gall's Law:

> “A complex system that works is invariably found to have evolved from a simple system that worked."

I would suggest: let's start with something people actually want to use, and move forward from there. First the QR codes and email notifications, we can always add the restocking and stuff later when we find out we really need it.

website/tampon/models.py
JobDoesburg: It was mainly related to the point I made earlier about whether we need to keep track of this.

website/tampon/forms.py
```diff
@@ -0,0 +1,72 @@
...
+            self.fields[f"stock_{stock.id}"] = forms.IntegerField(
+                label=stock.name,
+                initial=stock.restock_default,
+                widget=forms.NumberInput(
+                    attrs={
+                        "class": "form-control",
+                        "placeholder": f"Enter amount for {stock.name}",
+                    }
+                ),
+            )
```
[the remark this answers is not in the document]
Luukosaurus: Stock is only used for data analysis. If this stock goes below 0, then that means people filled it in wrong or did not restock it in the admin. If this happens, it's annoying for the purpose of data analysis, but it should not hinder resolving the notification. The min_value will be implemented.

website/tampon/models.py
Luukosaurus: Okay then yes, we use it if we decide that we want to keep track of this and as far as I was told, this was preferred by the committee. But I will come back to you with an explanation by the committee about the data collection and its analysis.

website/tampon/models.py
JansenThomas: We want to keep track of the data because what we are doing is not really liked by campus and facilities, this is just paid for by Faculty Board. So in the case that we need to defend what we are doing we really want to be able to show that it is actively being used. The amount of times we restock and how much seems like a good way to show this.

website/tampon/templates/tampon/index.html
Luukosaurus: This just adds a circle, or am I missing something

website/tampon/models.py
Luukosaurus: This part is mainly to transform 2 to 002 or 02 not to cut off 9999 to 999 or something like that.
I agree that the db can have restrictions on this, but it should not be needed because it's done once when a room is added to the system and then never looked at again. 
But this code will be looked at when I will be looking at removing room_code.

website/tampon/templates/tampon/index.html
JobDoesburg: Whoops, I meant `fa-droplet` hahah

website/tampon/models.py
JobDoesburg: Sure, but wouldnt just counting the amount of notifications also work? Do you really need detailed stock info?

It's fine by me to implement this if you really want it. Just warning you for a common pitfall of software engineering.

website/tampon/permissions.py
KiOui: The upside of Django is that you get an entire admin dashboard "for free" without having to program anything. You also get permissions for the models you create by default.

The functionality that is needed for this application (viewing, and resolving notifications) can both be done in the admin dashboard by default. Therefore I think it is a very good idea not to reinvent the wheel and just use the admin dashboard, and Django created permissions, for this use case.

Additionally, the `staff` flag only determines whether you can access the `/admin` dashboard. If you only have the `staff` flag set, you can still not do anything on the `/admin` dashboard because you don't have the proper permissions.

I therefore recommend to just use the admin dashboard and create a group in TOSTI (after this is merged) for all the menstruacie committee members. The group can then have permissions to access all models of the `tampon` application.

website/tampon/models.py
KiOui: I agree with @JobDoesburg here, let's do this.

KiOui:
> > Also: The idea is to hang up posters in the bathrooms with QR codes that people can scan right? Would it be nice to make a QR code in the admin backend of a room that can be printed out?
> 
> That really sounds like feature creep 🙂

Nooo why so? We can pretty easily make a QR code export button in the admin right that can then immediately be printed out? Then the QR codes don't need to be manually created for each room.

website/tampon/models.py
KiOui: Maybe something like `comment` is a more suiting name then.

JobDoesburg:
> > > Also: The idea is to hang up posters in the bathrooms with QR codes that people can scan right? Would it be nice to make a QR code in the admin backend of a room that can be printed out?
> > 
> > 
> > That really sounds like feature creep 🙂
> 
> Nooo why so? We can pretty easily make a QR code export button in the admin right that can then immediately be printed out? Then the QR codes don't need to be manually created for each room.


You're creating rooms just once, and then you're never using this functionality ever again, while there are dozens of QR code generators on the internet that work just fine.

Luukosaurus:
> > > > Also: The idea is to hang up posters in the bathrooms with QR codes that people can scan right? Would it be nice to make a QR code in the admin backend of a room that can be printed out?
> > > 
> > > 
> > > That really sounds like feature creep 🙂
> > 
> > 
> > Nooo why so? We can pretty easily make a QR code export button in the admin right that can then immediately be printed out? Then the QR codes don't need to be manually created for each room.
> 
> You're creating rooms just once, and then you're never using this functionality ever again, while there are dozens of QR code generators on the internet that work just fine.

We are not going to do this in the admin because this is not easier than doing it externally. We now have a script that creates the qr codes in a nicely formatted sticker, which works well.

website/tampon/models.py
Luukosaurus: Yea I agree. I will change this.

website/tampon/models.py
Luukosaurus: I was informed that, indeed, it also needs support for adding letters like a or b at the end, and this does not work in the current system. I then agree that changing this to a slug becomes beneficial, especially if there is again room formats that I was notified of, like hallways.

Luukosaurus: I'm planning on keeping the admin similar so that people can still fill it in by selecting a floor and a room number. And then also an option to directly set the slug/name. Do you think this is useful, or does this not make a lot of sense?
I can also make the slug based on the name or vice versa. 
My goal here is to make it less error-prone.

If i do this can i just not save eather the slug or the name in the database and just use the __str__ function so its only a pk and one string.

What are your thoughts about this?

website/tampon/templates/tampon/index.html
Luukosaurus: Yea i had this before, but I was told that it was indeed a bit too much.

website/tampon/views.py
```diff
@@ -0,0 +1,128 @@
...
+                messages.success(
+                    request,
+                    f"Notification for {notification.room} resolved and restock recorded.",
+                )
+        else:
+            messages.error(
+                request,
+                "Invalid input. Please enter a valid quantity to resolve the notification.",
+            )
+        return redirect("tampon:notifications")
```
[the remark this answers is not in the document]
Luukosaurus: @JobDoesburg @KiOui 
Do we need to actually make a test because this will take a bit.

website/tampon/permissions.py
Luukosaurus: I know how the admin works and its benefits, and I agree that this should be possible in a good enough way, but this was a design choice made early on. I requested a statement from the committee about why this is preferred this way or that is does not matter. But it most likely comes down to it being easier to access and use by committee members.

website/tampon/views.py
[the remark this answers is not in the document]
JobDoesburg: Nah

website/tampon/models.py
JobDoesburg: I dont think a filter/select is required personally. We're not talking about 10+ rooms I think. And its only the admins, not the users that make notifications. So I wouldnt put in too much effort for something that only very few people will use very rarely

website/tampon/permissions.py
WidadMajdoubi24: We discussed it with olympus and wanted to keep the interface like the "borrel reservations" for example, easy accessible from the page immediately for the less tech literate people in the committee. I think it does not really hinder, might be redundant, but might make it easier for those people. If you guys really disagree with it then we can remove it, but imo it prevents us from explaining anything to committee members since it is immediately visible.

website/tampon/models.py
Luukosaurus: Why is it bad to do it this way.
And the sel.is_resolved = true then has to be done externally.

website/tampon/models.py
Luukosaurus: I kept it simple, but since this is a big change, I probably need a new review. Especially since I used some new things I have never used before (annotated_rooms).

website/tampon/models.py
Luukosaurus: This is changed but please check if I missed somthing

website/tampon/migrations/0004_rename_notification_text_tamponnotification_comment.py
KiOui, line 1: Could you squash the migrations before this is merged? Just deleting the migration and running `makemigrations` will work fine 😄

website/tampon/admin.py
```diff
@@ -0,0 +1,58 @@
...
+        "name",
+        "slug",
+        "active",
+    )
+    list_filter = ("active",)
+    search_fields = (
+        "name",
+        "slug",
+    )
+
```
KiOui, line 25:
```suggestion
    prepopulated_fields = {"slug": ("name",)}
```

This will automatically fill in the `slug` for you 😄

website/tampon/models.py
KiOui: It's not bad to do it in this way, it's nicer to do this in the Django `save()` method as that automatically runs every time you save a model. This way you will never have to call `mark_resolved` but you can just set `is_resolved` and the model will automatically (always) set `resolved_at` to the moment you want when saving.

website/tampon/models.py
```diff
@@ -0,0 +1,86 @@
...
+        return f"Notification for {self.room} at {self.created_at}"
+
+
+class Restock(models.Model):
+    """Restock record for menstrual products."""
+
+    room = models.ForeignKey(Room, on_delete=models.PROTECT)
+    restock_time = models.DateTimeField(auto_now_add=True)
+    restocked_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True
```
KiOui: You should user `User` here and put a `User = get_user_model()` at the top of this file (if you search for that you can see other occurrences).


```suggestion
       User, on_delete=models.SET_NULL, null=True
```

website/tampon/permissions.py
KiOui: No if you want it this way that's fine, I'm just making suggestions!

Might be nice to put the `Notifications` button somewhere else because now it's in the header of a card.

website/tampon/models.py
```diff
@@ -0,0 +1,86 @@
...
+        verbose_name_plural = "Restock records"
+
+    def __str__(self):
+        return f"Restock for {self.room} at {self.restock_time}"
+
+
+class StockData(models.Model):
+    name = models.CharField(max_length=100)
+    restock_default = models.IntegerField(default=0)
+    stock_amount = models.IntegerField(default=0)
```
KiOui, line 77: What's this used for? Tracking how much of the product is in stock? If so, does this automatically update when I do a restock (or shouldn't it)?

website/tampon/permissions.py
```diff
@@ -0,0 +1,17 @@
...
+        if not is_tampon_committee_member(request.user):
+            raise PermissionDenied
+        return super().dispatch(request, *args, **kwargs)
+
+
+def is_tampon_committee_member(user):
+    """Return whether the user is allowed to see tampon notifications."""
+    return user.is_authenticated and (
+        user.is_staff or user.has_perm("tampon.manage_tamponnotification")
+    )
```
KiOui: Are staff users always allowed, even if they don't have any other permissions than `is_staff`?

website/tampon/views.py
```diff
@@ -0,0 +1,128 @@
...
+                self.request,
+                f"You are submitting a notification for {room}.",
+                extra_tags="TempMessage",
+            )
+        elif room_input:
+            messages.error(
+                self.request,
+                "Something went wrong with the room number provided. Please select a valid room from the dropdown menu.",
+                extra_tags="TempMessage",
+            )
```
KiOui, line 45: Suggestion for UI improvement (don't have to do this here, but maybe later):

The notification now appears at the top of the screen, might be nice to let it appear in the card instead and grey-out the room selector.

website/tampon/views.py
```diff
@@ -0,0 +1,128 @@
...
+            for field_name, quantity in form.cleaned_data.items():
+                if field_name.startswith("stock_"):
+                    stock_id = int(field_name.split("_")[1])
+                    stock_data = get_object_or_404(StockData, pk=stock_id)
+                    RestockItem.objects.create(
+                        restock=restock,
+                        stock_data=stock_data,
+                        quantity=quantity,
+                    )
+                    stock_data.stock_amount -= quantity
```
KiOui, line 113: Should this be able to be negative?

website/tampon/models.py
```diff
@@ -0,0 +1,86 @@
...
+        self.is_resolved = True
+        if self.resolved_at is None:
+            self.resolved_at = timezone.now()
+        self.save(update_fields=["is_resolved", "resolved_at"])
+
+    def __str__(self):
+        return f"Notification for {self.room} at {self.created_at}"
+
+
+class Restock(models.Model):
```
KiOui, line 48: I've now read all the code and see how this works. I don't think it's easily distinguishable from the name what these models for restocking and stock do. Could you add some in-code comments about their relationship (maybe at the top of each class)?

website/tampon/templates/tampon/notifications.html
```diff
@@ -0,0 +1,85 @@
+{% extends "tosti/base.html" %}
+{% block title %}T.A.M.P.O.N. notifications{% endblock %}
+{% block page %}
+<div class="container mt-5">
+    <div class="d-flex justify-content-between align-items-center mb-4">
+        <h1 class="h3 mb-0">T.A.M.P.O.N. notifications</h1>
```
KiOui: Could you keep the same TOSTI styling for this page? So with an H1 header and stuff (the thing you also did in the other page).

website/tampon/templates/tampon/index.html
KiOui, line 1: Could you add/remove some margins and maybe move the notification button out of the card?

Image

KiOui (changes requested):
Okay now I've done a full review. Thanks again for the good work! Most of it are suggestions 😄

website/tampon/models.py
KiOui: Nvm, I just saw that it does.

website/tampon/admin.py
Luukosaurus: Yea thats perfect. I did not know that existed. I first wanted to remove the slug part entirely and do it on save to make this easier, but this is way better.

website/tampon/views.py
Luukosaurus: Yes, if the committee does not make mistakes, it should never go to negative, but if they do, I don't want to block anything. And if they then again restock it in the backend, they can just add 1000 and don't have to think about how much stock was voided since it can't go negative and thus no information is lost.

website/tampon/templates/tampon/notifications.html
Luukosaurus: I did not make most of this part besides the form popup and thus its backend functionality, so I did not want to change too much of how it looked, but I think this is more in line and should be enough, especially since it is only seen by committee members.

I also use a different formatter or a formatter at all than the person who last edited the layout, so it might look weird in the commit. 

(Do you know if I'm using the right formater because I could only find the Python (Black) formater?) .

website/tampon/templates/tampon/index.html
Luukosaurus: I left the notification in the card because it does not affect the normal card, only for committee members, and it's above the original card-header. It is now also prominent on the page for the people who most likely need to go there and not use this page at all.

If you have an idea how I can do this instead, I'm open to suggestions.

Luukosaurus:
Is this now good because I lost track of what things should be changed and what things don't need change?

KiOui:
> Is this now good because I lost track of what things should be changed and what things don't need change?

Hey @Luukosaurus! You need to answer/resolve all open threads. After that request another review, when everything is good one of us will approve.

website/tampon/permissions.py
Luukosaurus: GitHub hid this comment from me, but indeed, this is most likely not the idea, so I will fix this.

JobDoesburg (approved):
Code looks good to me. I think the UI can have some improvements maybe, but we can fix them later (I might make my own proposal). If you fix the linting, we can merge imo. 

Let's evaluate after a few weeks if the whole restocking is actually being used, or that we may need to change some stuff about it.

Luukosaurus:
Im unable to run linting locally for flake8, and flake8 is also new to me. I'm used to ruff and vscode, and I feel like this project is made for something else, like PyCharm, due to the absence of .vscode files. I will probably now just edit it manually. I am also unable to run the workflow myself, but I guess this is a deliberate choice or that I just can't find it. 

The readme.md says "uv run flake8 website", but this finds issues in almost every file. This is also different than what the GitHub action does since this one has a max length way higher. This again can be me not knowing how flake8 works or me overlooking something.

PR #690, opened 2026-08-31
Fix queue endpoint crash and polling JS errors for unavailable players

JobDoesburg:
Follow-up to #689. Deploying that PR surfaced a pre-existing bug: `SpotifyPlayer.queue` had its `None` check *after* the list comprehension that subscripts the Spotify response — dead code while `do_spotify_request` still raised on auth errors, a crash (TOSTI-YY, `TypeError: 'NoneType' object is not subscriptable`) now that it returns `None` for the revoked token.

- `SpotifyPlayer.queue`: check for `None` before building the track list; cache "unavailable" as before.
- `SpotifyPlayer.request_song`: raise `SpotifyException` when the track lookup returns `None` — the API view already maps that to a 503 and the MCP layer to an error payload — instead of crashing later in `SpotifyQueueItem.queue_track`.
- `queue.html` / `requests.html`: the 5-second polling loops called `response.json()` without checking `response.ok`, so any 500 produced an unhandled `SyntaxError: Unexpected token '<'` in the console every poll (reproduced live on `/thaliedje/player/noord/`). They now check `response.ok`, tolerate a `null` queue (what the fixed endpoint returns for an unavailable player), and keep the last known state on failure. `player.html`'s control buttons already handled errors with a toast; `search` already guarded `None` server-side.

Tests pass. Reminder: the Thaliedje Noord Spotify account still needs manual re-authorization via the admin — that's the root cause of the degraded state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01B9SHJqHKhHkdE7YpxApR55


JobDoesburg:
Added a second commit: control actions (play/pause/skip/volume/shuffle/repeat/queue track) now fail loudly with a 503 + friendly `detail` message instead of silently returning 200 when Spotify is unavailable, and the player UI shows a dedicated "The music player is currently unavailable" toast on 503. Read/display paths still degrade quietly. The best-effort auto-start after queueing a track keeps its deliberate silent behavior.

